PR #23, opened 2021-11-07
fix: read without -r is generally bad.

docwhat:
The yes/no prompt doesn't need backslash escaped characters.

From z-shell/zinit#3


pschmitt:
Thanks!

PR #24, opened 2021-11-07
fix: use [*] inside arbitrary strings.

docwhat:
Only use `[@]` as only content of a string. e.g., "${foo[@]}"

From z-shell/zinit#5


pschmitt:
yup, thanks!

PR #25, opened 2021-11-08
fix: package are broken again

docwhat:
You cannot refer to a variable in the same statement as it is declared.


pschmitt:
Right. Thanks for that. We most def need some automated testing.

PR #39, opened 2021-11-10
No more name calling

pschmitt:
I don't think its fair to have psprint's real name hilighted so prominently in
the disclaimer.

Sure, we can link to his profile - if he decides to put his real name on there
that's his choice. Anyway, we shouldn't make this decision for him. We
aren't at war with the poor guy.


alichtman:
This was not the intent of the note when I put it there, however, I can see how it would be taken that way. Thanks for flagging.

Issue #36, opened 2021-11-09
🧹 Repo clean up


pschmitt:
A few things need to be removed or altered IMHO:

- [x] To be removed: `.mailmap`
- [x] To be moved: `zsdoc/` to `doc/`
- [x] ~~To be renamed: `doc/` -> `docs/`~~
- [x] we should get rid of .travis.yml and run the test with Github Actions
- [x] the tests should live in the main/master branch (they are currently on a branch: https://github.com/zdharma-continuum/zinit/tree/tests)
- [x] the source of gh-pages should live in the main/master branch (they are currently on a branch: https://github.com/zdharma-continuum/zinit/tree/documentation)
- [x] Rename master to main #46

PR #42, opened 2021-11-11
CI

pschmitt:
Current state:

- A whole lof of containers are getting built! Their main purpose is to quickly test zinit against (older) zsh versions. So they essentially contain zsh and zinit. There are a few helper functions available inside it (`zinit::<tab>`). I've mainly focussed on the alpine-based container for now.
- Tests are getting run. But none of them pass... They all need to be updated since they were written in the zplugin days.
- Docs are getting generated and the build is set to fail if there are not up to date. This isn't exactly working as expected though since the generated docs seem to not be generated in reproducible way (To be debugged further)

Here's what is interesting to look at: 
- CI logs: https://github.com/pschmitt/zinit/actions
- Containers that are being built: https://github.com/pschmitt/zinit/pkgs/container/zinit 
- `docker run -it --rm ghcr.io/pschmitt/zinit:latest`

Refs:
- #36
- #37
- https://github.com/zdharma-continuum/zinit/discussions/21
- https://github.com/zdharma-continuum/zinit/blob/bef57e843d0d8c239fcb208cc0576216c683f788/.travis.yml#L26-L35


pschmitt:
Superseded (partially) by #70

pschmitt:
Partially superseded by #117 as well.

pschmitt:
The only feature that's left here that's not already in the main branch is the "make test" bit that's broken (the tests are, not the CI part of it)

vladdoster:
Closing due to lack of activity.

PR #47, opened 2021-11-11
Update annexes repo names

pschmitt:
See https://github.com/zdharma-continuum/zinit/issues/38

~~DO NOT MERGE YET. The repos have not been renamed yet.~~ DONE!

PR #48, opened 2021-11-11
fix: more cleaning up urls

docwhat:


pschmitt:
#47 😉

docwhat:
Okay, rebuilt this off PR #47... you should be able to just apply both.

PR #49, opened 2021-11-11
add editorconfig to help consistency

docwhat:
https://editorconfig.org FTW!

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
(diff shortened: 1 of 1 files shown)


pschmitt:
Thanks, I really hate it.

Rather than storing extra metadata in yet another config dir that many tools ignore I suggest you teach your editor how to read [modelines](https://vim.fandom.com/wiki/Modeline_magic)

Every single source file has one at the bottom.

```shell
grep -R 'vim:' .
```

pschmitt (changes requested):
It's a viiiiiiiiiiiiim's world! 🎼

Issue #45, opened 2021-11-11
💡 Proposal: Move doc/install.sh to another dir


pschmitt:
While working on #43 I couldn't help but notice that [the installer script](./doc/install.sh) is stored under `doc/`.

IMHO the `doc/` dir should only - as the name implies - store documentation and not arbitrary scripts.

I'd like to move it to `installer/install.sh` or `scripts/install/install.sh`.
This shouldn't have many implications besides that the `curl` install instructions need to be updated once (and if) this is done.

Related: #55

pschmitt:
Implemented in #61

PR #61, opened 2021-11-14
Overengineer the installer

### LINKED ISSUE #60: Rewrite the installer
This is a follow up to #43 

Installer wishlist:

- [x] Functions, everywhere!
- [x] No more color escapes sequences in the echo calls (but vars)
- [ ] Optionally: Ask to install [zsh-bin](https://github.com/romkatv/zsh-bin) if no zsh installed (or zsh version < 5.5)
- [x] #45 ?

pschmitt:
Fixes #60 and #45

TODO:

- [x] more emojis
- [x] warn about zsh version if below 5.5.1
- [x] Shellcheck in ci
- [x] Update CHANGELOG
- [x] Cygwin jobs
- [x] macOS jobs (should work ootb, but they hang indefinitely as of 2021.11.15)


alichtman:
Let's move install.sh out of doc

pschmitt:
> Let's move install.sh out of doc
Done. I went ahead and moved `doc/mod-install.sh` along with it.
They now both reside in `scripts/`.

README.md
```diff
@@ -963,7 +963,7 @@ helper functions (`zicompinit`,`zicdreplay` & `zicdclear` – see below for expl
 of the last one). To summarize:
 
 ```zsh
-source ~/.zinit/bin/zinit.zsh
+source ~/.local/share/zinit/zinit.git/zinit.zsh
```
alichtman, line 966: Can we XDG-ify this?

scripts/install.sh
```diff
@@ -0,0 +1,338 @@
...
+# Load a few important annexes, without Turbo
+# (this is currently required for annexes)
+zinit light-mode for \\
+    zdharma-continuum/zinit-annex-as-monitor \\
+    zdharma-continuum/zinit-annex-bin-gem-node \\
+    zdharma-continuum/zinit-annex-patch-dl \\
+    zdharma-continuum/zinit-annex-rust
+
+EOF
+  # Ask user if we should add the annexes to his zshrc
```
alichtman, line 244:
```suggestion
  # Ask user if we should add the annexes to their zshrc
```

scripts/install.sh
```diff
@@ -0,0 +1,338 @@
...
+  fi
+
+  command cat <<-EOF >> "$ZSHRC"
+### End of Zinit's installer chunk
+EOF
+  unset reply zshrc_annex_file
+}
+
+
+display_tutorial() {
```
alichtman, line 276: Let's drop this and just point the user to the wiki for a tutorial

alichtman (changes requested):

README.md
pschmitt: Done.

scripts/install.sh
pschmitt: Good idea. The sad is that I just rewrote some part of it the "tutorial". Anyway.

pschmitt: -> 0c425c6

pschmitt:
FYI: I submarined a ZPLG -> ZINIT sed and replace in the manpage.
428d71822986fcca0575ef637387aee01f53842f

pschmitt:
The only thing left here is to pray that in the future the macOS runners work as they should 😕

doc/CHANGELOG.md
```diff
@@ -9,6 +9,23 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+* 16-11-2021
+  - A brand-new installer has been developped. A few new features have been
```
alichtman, line 13:
```suggestion
  - A brand-new installer has been developed. A few new features have been
```

pschmitt: :+1:

scripts/install.sh
```diff
@@ -0,0 +1,331 @@
...
+         "- the most useful annexes (Zinit extensions that add new" \
+         "functions-features to the plugin manager) to the zshrc as well?" \
+         "It will be the following snippet:"
+    command cat "$zshrc_annex_file"
+    # shellcheck disable=2059
+    printf "${COLOR_PALE_MAGENTA}${COLOR_RESET} Enter y/n and press Return: "
+    read -r reply
+  fi
+
+  if [ "$reply" = y ] || [ "$reply" = Y ]; then
```
alichtman, line 265:
```suggestion
  if [[ "$reply" =~ ^(Y|y)$ ]]; then
```

alichtman: https://stackoverflow.com/a/3232082

scripts/install.sh
```diff
@@ -0,0 +1,331 @@
+#!/usr/bin/env sh
```
alichtman, line 1:
```suggestion
#!/usr/bin/env bash
```

alichtman: We don't need to make this `sh` compatible :)

alichtman (changes requested):

scripts/install.sh
pschmitt: oh, woops..! I may have pressed enter too quickly when prompted with my snippet.

pschmitt: Erm, this is and always has been a POSIX sh script.
I put in a lot of effort to make it posix sh compliant, I'm don't think this is a good idea.

pschmitt: Also since this is a zsh plugin manager for me it does not make sense to require bash to be installed to get it. It would only hurt users that are trying to get zinit.

alichtman: > Also since this is a zsh plugin manager for me it does not make sense to require bash to be installed to get it. It would only hurt users that are trying to get zinit.

_Most_ OSes come with bash installed already (and if you're on an OS that doesn't, you certainly know how to install it). I see POSIX `sh` compliance as a hindrance to development here, but since you're doing the work, it's really up to you,

pschmitt: Most do, but not all. That's precisely the point of using posix instead of bash.
alpine, busybox etc do not ship bash for instance. 

Requiring bash now that is always worked with pure posix sh will also break some workflows.
I don't want to break any existing workflows.

pschmitt: As to further development I don't see POSIX being a hindrance. The installer is pretty static, for me it is pretty much feature-complete already.

pschmitt:
GH macOS Runners seem to be fixed now. Enabled macOS jobs.

scripts/install.sh
pschmitt: That's 2 bash-isms

shellcheck will call you out on that:

- https://github.com/koalaman/shellcheck/wiki/SC3010
- https://github.com/koalaman/shellcheck/wiki/SC3015

PR #67, opened 2021-11-17
(bugfix): add macos as possible release descriptor

vladdoster:
Fixes issue where Neovim uses `macos` as a descriptor and fails to download.


pschmitt:
LGTM

PR #78, opened 2021-11-18
Print the correct module path

joakimgottzen:
https://github.com/zdharma-continuum/zinit-module/commit/a1125ba updated the path of the compiled module, but it's not reflected in the output after a successful build by 'zinit module build'.

Update the .zinit-module function to print the correct module path.


pschmitt:
Good catch. Thanks!

PR #82, opened 2021-11-18
🧹 Move template plugin data to share/template-plugin

pschmitt:
The template files used by `zinit create` used to live under `doc/`

This moves them to a more sensible location: `share/template-plugin`
Should make a bit more sense now, I for sure was confused about what that was and had to dig though the code to figure it out :)

⚠️ This introduces 2 minor changes:
- newly created plugins are no longer MIT licensed by default. That should really be the users' decision, not ours.
- new plugin repos won't contain a default `.gitattributes`

Refs:
- #36


alichtman:
> newly created plugins are no longer MIT licensed by default. That should really be the users' decision, not ours.

Default is king. If the user wants to change it, they will. Otherwise, I foresee many people releasing projects without a license.

I vote for keeping MIT as the default license.

zinit-autoload.zsh
```diff
@@ -3037,22 +3037,20 @@ EOF
     xfuncname = "^#+[[:blank:]].*$"
 EOF
 
-    builtin print -r -- "*.zsh  diff=zsh" >! .gitattributes
-    builtin print -r -- "*.md   diff=markdown" >! .gitattributes
     builtin print -r -- "# $plugin" >! "README.md"
-    command cp -vf "${ZINIT[BIN_DIR]}/LICENSE" LICENSE
-    command cp -vf "${ZINIT[BIN_DIR]}/doc/Zsh.gitignore" .gitignore
-    command cp -vf "${ZINIT[BIN_DIR]}/doc/example-script" .
+    command cp -vf "${ZINIT[BIN_DIR]}/share/template-plugin/zsh.gitignore" .gitignore
```
alichtman, line 3041:
```suggestion
    command cp -vf "${ZINIT[BIN_DIR]}/LICENSE" LICENSE
    command cp -vf "${ZINIT[BIN_DIR]}/share/template-plugin/zsh.gitignore" .gitignore
```

alichtman (changes requested):
See above

pschmitt:
Kay, I partially reverted the contention change.
The MIT license file is placed in the repo by default again. I've updated the message string a bit.

PR #86, opened 2021-11-19
Fix zinit module build when module repo has already been cloned before

pschmitt:
Right now, the *second* call to `zinit module build`leads to:

```zsh
$ zi module build
fatal: destination path '/home/pschmitt/.local/share/zinit/module' already exists and is not an empty directory.
Failed to clone module repo
```

This PR addresses this by clearing the local repo and pulling from the remote if the module repo is already on disk.

### DIFF
--- a/zinit-autoload.zsh
+++ b/zinit-autoload.zsh
@@ -3330,16 +3330,16 @@ EOF
 .zinit-build-module() {
     setopt localoptions localtraps
     trap 'return 1' INT TERM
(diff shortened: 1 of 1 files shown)


pschmitt:
Once this is merged, #87 can be too.

alichtman:
Seems reasonable.

pschmitt:
Will merge #87 then!

PR #95, opened 2021-11-20
XDG compliance: `ZINIT[HOME_DIR]` no longer defaults to `HOME/.zinit` but to `XDG_DATA_HOME/zinit`

pschmitt:
We updated the installer to install to `XDG_DATA_HOME/zinit`, but never updated zinit itself to actually look for its home dir in there.

Note that this also updates the default value `ZINIT[HOME_DIR]` if none of these paths exists to `XDG_DATA_HOME/zinit` (instead of `HOME/.zinit`).
Epic victory royale for the XDG gang.

Refs:
- #61

### DIFF
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -9,6 +9,17 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+* 20-11-2021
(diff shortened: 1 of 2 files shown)


alichtman:
I patched this manually on my system last night and hadn't yet put up a PR. Thanks! XDG FTW

alichtman:
We should add zinit to the Arch wiki list of XDG-compliant programs

PR #99, opened 2021-11-21
Make sure TERM is set in the container. Fixes #97

### LINKED ISSUE #97: 🤔 Some ices get ignored when run non-interactively
**Describe the bug**
When running non-interactively, for eg with Ansible, some ices get ignored. This also happens when run with `zsh -sli`.
My suspicion is that there's something fishy going with zsh hooks (`add-zsh-hook`) in this scenario, since this is how the make ice is called.

This makes it impossible to install *everything* in one go.

**Zinit config**

```zsh
# make does not get executed when running `@zinit-scheduler-burst` w/o a TTY attached
zinit light-mode as"program" make for @direnv/direnv
```

Refs:
- https://github.com/zdharma-continuum/zinit/runs/4275129417?check_suite_focus=true
- #96

pschmitt:
Should fix the plugin zunit test :)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,9 +4,11 @@ FROM alpine:${VERSION}
 ARG PUSERNAME=user01
 ARG PUID=1000
 ARG PGID=1000
+ARG TERM=xterm-256color
 ARG ZINIT_ZSH_VERSION
+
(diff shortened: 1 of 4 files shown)


pschmitt:
Yup, That was it. Merging this one in since it's a no-brainer! #makeThePipelineGreenAgain

pschmitt:
image

PR #103, opened 2021-11-22
🐳 Add annexes to container

pschmitt:
Here are a few updates to the docker container.

- annexes get installed by default
- zinit completion
- at container build time `@zinit-scheduler burst` now installs to `/data-static`
- at runtime `/data-static` gets copied over to `/data`. This is necessary because `/data` is meant to be backed via a volume (that's what we do in the zunit tests for instance to verify if we downloaded the files correctly)
- most of the custom stuff (keys setup, annexes, the /data-static to /data file-copy etc) can now be disabled with `NOTHING_FANCY=1` at runtime

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -11,7 +11,7 @@ ENV PUSERNAME=${PUSERNAME} PUID=${PUID} PGID=${PGID} \
      TERM=${TERM} ZINIT_ZSH_VERSION=${ZINIT_ZSH_VERSION}
 
 RUN apk --no-cache --virtual base add \
-        zsh curl git libuser sudo && \
+        zsh curl git libuser sudo rsync && \
     apk --no-cache --virtual build-tools add \
         build-base autoconf ncurses-dev bash go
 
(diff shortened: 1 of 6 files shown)


alichtman:
Did this break something with zsh 5.1?

pschmitt:
Nah, that's just GH runners getting stuck.

pschmitt:
Seems related: https://github.com/zsh-users/zsh-docker/issues/21

pschmitt:
zsh 5.1.1 is for sure a trouble-maker, its build log ends with "build failed":

https://github.com/zdharma-continuum/zinit/runs/4292608375?check_suite_focus=true#step:7:6245

The image kinda works though:

```zsh
docker run -it --rm ghcr.io/pschmitt/zinit:zsh-5.1.1 zsh --version
zsh 5.1.1 (x86_64-unknown-linux-gnu)
```

Maybe we should drop zsh 5.1.1 altogether. zinit works best with 5.5.1+ anyway.

alichtman:
Yeah, let's drop it.

pschmitt:
Done: https://github.com/zdharma-continuum/zinit/commit/888153190453cc67af34c54f59f47ceb5372841f

PR #104, opened 2021-11-22
Fetch the annexes during the install

pschmitt:
This installs the annexes with `@zinit-scheduler burst` at install time, rather than when the user first starts zsh after the installation.


alichtman (approved):
Seems cool to me :) I do kinda like watching the install logs roll by on a fresh zinit start but I see the appeal of this

Issue #58, opened 2021-11-13
Remove all unicode emojis / symbols from file names


pschmitt:
**Describe the bug**
Zinit cannot compile annex since some of the handler files have unicode chars in their names which `zcompile` [does not support](https://www.zsh.org/mla/workers/2020/msg01057.html).
I wonder if we can dance around this issue by making the `compile` ice dynamically rename files.

**Zinit config**
```zsh
zinit light-mode compile'*handler' for \
     zdharma-continuum/zinit-annex-bin-gem-node
```

**Error output**

```
(anon):zcompile:2: can't open file: /home/pschmitt/.local/share/zinit/plugins/zdharma-continuum---zinit-annex-bin-gem-node/\M-b\M-^F\M-^Rza-bgn-atclone-handler                                         
(anon):zcompile:2: can't open file: /home/pschmitt/.local/share/zinit/plugins/zdharma-continuum---zinit-annex-bin-gem-node/\M-b\M-^F\M-^Rza-bgn-atdelete-handler                                        
(anon):zcompile:2: can't open file: /home/pschmitt/.local/share/zinit/plugins/zdharma-continuum---zinit-annex-bin-gem-node/\M-b\M-^F\M-^Rza-bgn-atload-handler                                          
(anon):zcompile:2: can't open file: /home/pschmitt/.local/share/zinit/plugins/zdharma-continuum---zinit-annex-bin-gem-node/\M-b\M-^F\M-^Rza-bgn-help-handler
```

vladdoster:
I don't think Unicode is the issue here

pschmitt:
Why? Got a theory on that?

vladdoster:
Did I think they already had Unicode in their name? Such as `zinit-annex-patch-dl`?

pschmitt:
> Did I think they already had Unicode in their name? Such as `zinit-annex-patch-dl`?

Most of them have one or more zsh files that contain a unicode arrow (→) in their name.

The obvious fix is to rename these files in the annex repos. This might be the cleaner way of handling this.

Is there anything preventing us from doing so? 🤔

alichtman:
I don’t think so. Happy to get rid of the arrow.

docwhat:
This is actually a known bug is zsh:  https://www.zsh.org/mla/workers/2020/msg01058.html

alichtman:
Regardless of the root cause, I think the right fix is to remove Unicode arrows from filenames. They cause more trouble than they're worth.

[...]

PR #107, opened 2021-11-25
Add bz2 extension to bz2 files before attempting to extract. Fixes #105

### LINKED ISSUE #105: 📦 ziextract should rename .bzip2 files before attempting to extract
While working on https://github.com/zdharma-continuum/zinit-packages/pull/2 I noticed that `ziextract` does detect bzip2 files correctly but ultimately fails to extract them when their suffix does *not* match `bz2|bzip2`.

This is actually a `bunzip2` issue, it refuses to operate on files whose extension don't match.

```
bunzip2: zinit-package-apr: unknown suffix - ignored
```

This is easily fixable by renaming the file just before attempting the extraction. The same is already done for zx packages.

Refs:
- See the `apr` package: https://github.com/zdharma-continuum/zinit-packages/blob/main/apr/package.json

pschmitt:


zinit-install.zsh
```diff
@@ -1683,6 +1683,13 @@ ziextract() {
...
             ;;
         ((#i)*.bz2|(#i)*.bzip2)
+            # Rename file if its extension does not match "bz2". bunzip2 refuses
+            # to operate on files that are not named correctly.
+            # See https://github.com/zdharma-continuum/zinit/issues/105
+            if [[ $file != (#i)*.bz2 ]] {
+                command mv $file $file.bz2
+                file=$file.bz2
+            }
             →zinit-extract() { →zinit-check bunzip2 "$file" || return 1
```
alichtman, line 1686: Looks like a missed Unicode char

alichtman (approved):

zinit-install.zsh
pschmitt: "missed" by whom?
This is a function name!

pschmitt: https://github.com/zdharma-continuum/zinit/issues/58 only covered file names

alichtman: Ah, gotcha. We should spin off a task for removing those from function names. I don't think we should have any unicode chars in the codebase except for strings we're printing for the user. It makes the code a lot harder to work with

alichtman: https://github.com/zdharma-continuum/zinit/issues/108

PR #112, opened 2021-11-26
Pedantic warning: Return != 0 when a hook fails

### LINKED ISSUE #111: :boom: `atclone`'s exit code shoud be propagated to the parent `zinit` call
zinit currently "silently" fails when the `atclone` script fails:

```zsh
zinit id-as"atclone-fails" atclone'echo atclone; false' for zdharma-continuum/null && echo okay || echo failed

Downloading zdharma-continuum/null... (at label: atclone-fails...)

Cloning into '/data/plugins/atclone-fails'...
▸▹▹▹▹ ███████████ OBJ: 100, , REC: 100%
No files for compilation found.
atclone
okay
```

The snippet above should echo `failed`.

pschmitt:
Fixes #111

This also logs an explicit warning when a hook fails:

```
Warning: ∞zinit-compile-plugin-hook hook returned with 1
```

### DIFF
--- a/zinit-install.zsh
+++ b/zinit-install.zsh
@@ -479,6 +479,11 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
             for key in "${reply[@]}"; do
                 arr=( "${(Q)${(z@)ZINIT_EXTS[$key]:-$ZINIT_EXTS2[$key]}[@]}" )
                 "${arr[5]}" plugin "$user" "$plugin" "$id_as" "$local_path" "${${key##(zinit|z-annex) hook:}%% <->}"
+                hook_rc=$?
+                [[ "$hook_rc" -ne 0 ]] && {
+                    rc="$hook_rc"
+                    builtin print -Pr -- "${ZINIT[col-warn]}Warning:%f%b ${ZINIT[col-obj]}${arr[5]}${ZINIT[col-warn]} hook returned with ${ZINIT[col-obj]}${hook_rc}${ZINIT[col-rst]}"
+                }
             done
         }
 
--- a/zinit-install.zsh
+++ b/zinit-install.zsh
@@ -2167,7 +2167,7 @@ zimv() {
     local extract=${ICE[extract]}
     @zinit-substitute extract
 
-    (( ${+ICE[extract]} )) || return
+    (( ${+ICE[extract]} )) || return 0
 
     .zinit-extract plugin "$extract" "$dir"
 }
--- a/zinit-install.zsh
+++ b/zinit-install.zsh
@@ -2085,12 +2085,13 @@ zimv() {
 
     local make=${ICE[make]}
     @zinit-substitute make
-    (( ${+ICE[make]} )) || return 0
+
+    (( ${+ICE[make]} )) || return
+    [[ $make = "!!"* ]] || return
 
     # Git-plugin make'' at download
-    [[ $make = "!!"* ]] && \
-        .zinit-countdown make && \
-            command make -C "$dir" ${(@s; ;)${make#\!\!}}
+    .zinit-countdown make && \
+        command make -C "$dir" ${(@s; ;)${make#\!\!}}
 }
 # ]]]
 # FUNCTION: ∞zinit-make-e-hook [[[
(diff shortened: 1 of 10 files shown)


pschmitt:
Surprise, surprise: this makes the zunit tests fail ;)

There's still a bit more work to do here.

pschmitt:
Okay, this was easier than I thought.

This is still a potentially breaking change, since it makes zinit quite a bit more strict in regards to return codes.

alichtman:
> This is still a potentially breaking change, since it makes zinit quite a bit more strict in regards to return codes.

I'm cool with a major version bump if this could change behavior of other programs. We've done some pretty heavy modification, so it's getting time for a major release anyways.

pschmitt:
> so it's getting time for a major release anyways.
Hence the v4.0 milestone :)

zinit-install.zsh
```diff
@@ -2229,19 +2262,20 @@ zimv() {
...
-                } else {
-                    .zinit-compile-plugin "$id_as" ""
-                }
+    if ! [[ ( $hook = *\!at(clone|pull)* && ${+ICE[nocompile]} -eq 0 ) || \
+            ( $hook = at(clone|pull)* && $ICE[nocompile] = '!' ) ]] {
+        return 0
+    }
+
+    # Compile plugin
+    if [[ -z $ICE[(i)(\!|)(sh|bash|ksh|csh)] ]] {
```
alichtman, line 2271: No idea what `$ICE[(i)(\!|)(sh|bash|ksh|csh)` evaluates to

alichtman (approved):
image

zinit-install.zsh
```diff
@@ -2100,10 +2108,12 @@ zimv() {
     local make=${ICE[make]}
     @zinit-substitute make
 
+    (( ${+ICE[make]} )) || return 0
```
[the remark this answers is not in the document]
pschmitt: No. That checks if "make" is in the ice assoc array.

zinit-install.zsh
```diff
@@ -2130,11 +2142,12 @@ zimv() {
...
+    [[ $make != "!"* ]] || return 0
```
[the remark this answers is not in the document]
pschmitt: In theory yes, but since you cannot really pass an array to a function this would be a very bad idea imo.

zinit-install.zsh
pschmitt: This looks for and returns the first key (ie `sh`, `bash`, `ksh` etc) of the `ICE`map that matches `(sh|bash|ksh|csh)` and excludes `!bash`, `!sh` etc.

pschmitt: ie this check if any of the `sh`, `bash`, `ksh` or `csh` ices is set.

alichtman: Insane. Thanks for the explanation.

pschmitt: Got it wrong btw. I've updated my comments.

pschmitt:
```zsh
typeset -A MYICE=([\!a]=valA [\!b]=valB [c]=C)
echo $MYICE[(i)(\!|)(a|b|c)]
!a
```

pschmitt:
This needs to a dedicated ci check.

```zsh
zinit load atclone'return 69' for zdharma-continuum/null
```
's RC should match `69` in zunit

pschmitt:
Done (see the ices zunit test)

Also: I've forwarded the rc even further. You get the actual != 0 rc of the failing ice now instead of just `1` when anything goes wrong.

pschmitt:
Update: `zinit update` return the right rcs now

pschmitt:
Forgot to mention: the `atpull` ice now has a dedicated zunit test and correctly forwards its rc to zinit

pschmitt:
Okay, docs updated and all. This is ready IMO.
Any last words?

flyxi121:
In `zinit-install.zsh` the Function `zinit-mv-hook` can never return 0 as one of 
```
command mv -vf "${afr[1]}" "$to"
command mv -vf "${afr[1]}".zwc "$to".zwc 2>/dev/null
```
will always return 1. 
So zinit always shows that an error happened on the `mv-hook` even though it ran successfully.

pschmitt:
Thanks for the hint. Will fix that.

PR #120, opened 2021-11-30
fix install script

### LINKED ISSUE #119: Wrong handling of ZDOTDIR in install script
**Describe the bug**
If `$ZDOTDIR` is set and `$ZSHRC` is not, the zinit installation script (`scripts/install.sh`) fails to update `.zshrc` with an error like:

```sh
sh: 214: cannot create /home/username/.zsh: Is a directory
```

**Zinit config**
`~/.zshenv`

```sh
export ZDOTDIR=$HOME/.zsh
```

`~/.zsh/.zshrc`

```sh
# empty
```

then run

```sh
sh -c "$(curl -fsSL https://git.io/zinit-install)"
```

**Expected behavior**
Update `$ZDOTDIR/.zshrc` without error

**Versions:**
 - $ZSH_VERSION == 5.8
 - Operating system name and version (uname -a): macOS, Linux (it is not related to operating system)

pickoba:
Currently, the zinit installation script (`scripts/install.sh`) generates the path to zshrc with the following code.

https://github.com/zdharma-continuum/zinit/blob/7f456fdbd419ebdaf68a2aa5b53fd6bd3a4edb48/scripts/install.sh#L308

If `$ZDOTDIR` is set and `$ZSHRC` is not, the right hand side will be evaluated as `$ZDOTDIR`. However, it should be `$ZDOTDIR/.zshrc`.

This PR fixes #119

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -305,7 +305,7 @@ ZINIT_COMMIT="${ZINIT_COMMIT:-}"  # no default value
 ZINIT_HOME="${ZINIT_HOME:-${XDG_DATA_HOME:-${HOME}/.local/share}/zinit}"
 ZINIT_REPO_DIR_NAME="${ZINIT_REPO_DIR_NAME:-zinit.git}"
 ZINIT_INSTALL_DIR=${ZINIT_INSTALL_DIR:-${ZINIT_HOME}/${ZINIT_REPO_DIR_NAME}}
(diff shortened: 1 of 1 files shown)


pschmitt:
Good catch!

pickoba:
Thank you for merging my first OSS PR! I'm really happy to become a contributor of this awesome project 😄

PR #135, opened 2021-12-10
Explicitly check availability of jq

daehyeok:
Explicitly check availability of jq on the .zinit-get-package.

### DIFF
--- a/doc/zsdoc/zinit-side.zsh.adoc
+++ b/doc/zsdoc/zinit-side.zsh.adoc
@@ -55,8 +55,8 @@ Called by:
 
  .zinit-exists-physically-message
  zinit-autoload.zsh/.zinit-clear-completions
- zinit-autoload.zsh/.zinit-compiled
  zinit-autoload.zsh/.zinit-compile-uncompile-all
+ zinit-autoload.zsh/.zinit-compiled
  zinit-autoload.zsh/.zinit-create
  zinit-autoload.zsh/.zinit-exists-message
  zinit-autoload.zsh/.zinit-get-completion-owner-uspl2col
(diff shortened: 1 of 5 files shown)


doc/zsdoc/zinit-side.zsh.adoc
```diff
@@ -138,12 +138,12 @@ Uses feature(s): _trap_
 Called by:
 
  zinit-autoload.zsh/.zinit-run-delete-hooks
- zinit-install.zsh/∞zinit-atclone-hook
- zinit-install.zsh/∞zinit-atpull-e-hook
- zinit-install.zsh/∞zinit-atpull-hook
- zinit-install.zsh/∞zinit-make-ee-hook
- zinit-install.zsh/∞zinit-make-e-hook
- zinit-install.zsh/∞zinit-make-hook
+ zinit-install.zsh/$'\342'$'\210'$'\236'zinit-atclone-hook
```
pschmitt, line 141: Odd. That shouldn't have happened.

daehyeok: hmm could you provide Dokerfile for `make doc` environment?

I use ubuntu:latest image and only install `git make zsh tree` through apt

daehyeok:
Note for why i want to add this code.

I found `zinit pack` command not print any warn/error message when jq not installed on systesm.
I don't have enough z-shell knowledge.  but it look first jq call is [zinit-install.zsh#L119](https://github.com/zdharma-continuum/zinit/blob/9ee45d6804fdf4972a9e913a1ad86d77de124231/zinit-install.zsh#L119) .  
`profiles=("${(@f)$(.zinit-json-get-value "$pkgjson" "${json_ices} | keys[]")}") || return 1`
which died silently.  Currently  `zinit pack` is only command which depend on `jq`. so i think explicitly add check routine to warn user when `jq` not installed. 

```
❯ .zinit-jq-check
❌ ERROR: jq binary not found 
Please install jq: https://github.com/stedolan/jq 
To do so with zinit, please refer to: https://github.com/zdharma-continuum/zinit/wiki/%F0%9F%A7%8A-Recommended-ices#jq
❯ profiles=("${(@f)$(.zinit-json-get-value "$pkgjson" "${json_ices} | keys[]")}") || return 1
   ~/workspace/zinit jq_warn ❯
```

 `profiles=("${(@f)$(.zinit-json-get-value "$pkgjson" "${json_ices} | keys[]")}") || return 1` 
and it silently die without print.

daehyeok:
@pschmitt  any update for doc test on CI?

PR #142, opened 2021-12-18
Add link ice for local-only snippets that uses symlink

twang817:
Has a narrow use case, but I added a new ice "link" for local snippets that uses symlinks instead of copying files into the cache.

For those of us that keep our entire dotfiles in a git repo (dotbare, et al), we often want to treat a lot of our own rc files as snippets that can be managed by zinit.  This is great for taking advantage of features such as `trigger-load`, `if`, or `has` ices.  I used to use local plugins for this, but noticed that the load times for plugins were significantly longer than snippets.

In migrating to snippets, I found that copying files from a source path into the ZINIT_SNIPPETS directory meant that I had to run `zinit update [snippet]` in order for any changes that I may make to my rc files to take effect.  This ice simply replaces the `cp` command with `ln`, thus changes were live instantly with a new shell.  I thought that snippets, in combination with symlinks, are an effective way to manage a lot of low-level rc file stuff that doesn't necessarily need to be in a separate git repository of its own.

I'm not fond of the ice name, as I think the word 'link' may be too general.  I also think that there may be some need for additional handling to ensure that this ice doesn't interfere with operation of normal URL-based snippets.  But, nonetheless, I thought I'd open this PR as a way to start discussing whether this could be worked to a point of merging into the mainline.

I am very new the zinit codebase, and I really had to flex my zsh to read through its code.   Please let me know if there are things that need to be adjusted in this PR.

### DIFF
--- a/doc/zsdoc/zinit-install.zsh.adoc
+++ b/doc/zsdoc/zinit-install.zsh.adoc
@@ -323,7 +323,7 @@ ____
  This is used to provide a layer of support for Oh-My-Zsh and Prezto.
 ____
 
-Has 357 line(s). Calls functions:
+Has 370 line(s). Calls functions:
 
  .zinit-download-snippet
  |-- zinit-side.zsh/.zinit-store-ices
--- a/zinit-autoload.zsh
+++ b/zinit-autoload.zsh
@@ -2212,22 +2212,26 @@ ZINIT[EXTENDED_GLOB]=""
         fi
         [[ -z $EPOCHREALTIME ]] && attime="<no zsh/datetime module → no time data>"
 
+        local line="$time"
         if [[ "$opt" = *-[a-z]#m[a-z]#* ]]; then
-            time="$attime"
+            line="$attime"
+        elif [[ "$opt" = *-[a-z]#a[a-z]#* ]]; then
+            line="$attime $line"
         fi
+
+        line="$line - $REPLY"
 
         if [[ ${sice[as]} == "command" ]]; then
-            builtin print "$time" - "$REPLY (command)"
+            line="$line (command)"
         elif [[ -n ${sice[sbin]+abc} ]]; then
-            builtin print "$time" - "$REPLY (sbin command)"
+            line="$line (sbin command)"
         elif [[ -n ${sice[fbin]+abc} ]]; then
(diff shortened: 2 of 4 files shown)


twang817:
Here are some examples:

```
zinit is-snippet for \
    id-as'aliases-general' link $DOTFILES/zsh/aliases/general.zsh \
    id-as'env'             link $DOTFILES/zsh/env/env.zsh \
    id-as'history'         link $DOTFILES/zsh/history/history.zsh \
    id-as'keybindings'     link $DOTFILES/zsh/keybindings/keybindings.zsh \
    id-as'options'         link $DOTFILES/zsh/options/options.zsh \
    ;
```

```
$ tree "${ZINIT[SNIPPETS_DIR]}"
/Users/Tommy.Wang/.dotfiles/zsh/snippets
├── OMZ::plugins
├── aliases-general
│   ├── ._zinit
│   │   ├── depth
│   │   ├── id-as
│   │   ├── is-snippet
│   │   ├── light-mode
│   │   ├── link
│   │   ├── lucid
│   │   ├── mode
│   │   ├── teleid
│   │   ├── url
│   │   └── wait
│   └── aliases-general -> /Users/Tommy.Wang/.dotfiles/zsh/aliases/general.zsh
...
└── plugins -> OMZ::plugins
```

```
$ zinit times
    2 ms - aliases-general
   13 ms - env
    2 ms - history
    2 ms - keybindings
    2 ms - options
```

For comparison, these load times were about 30ms each when loading them via local plugins.

vladdoster:
@twang817 Looks like there is one task needing your attention.

```shell
❌ The zshelldocs in the repo are not up to date
Please regenerate them with $ make doc
Error: Process completed with exit code 1.
```

vladdoster (changes requested):
```shell
❌ The zshelldocs in the repo are not up to date
Please regenerate them with $ make doc
Error: Process completed with exit code 1.
```

twang817:
I've been beating my head against this for a bit now.  Following HACKING.md, I have been running zsd in a Docker container (using zshuser/zsh:latest), but when I run `make doc`, I seem to get adocs that have quite a bit of differences.

twang817:
I goofed up by starting this PR w/ master.  Let me recreate the PR.

PR #143, opened 2021-12-21
More robust handling of symlinks in the completions folder

nbrown:
zinit clist assumes that all symlinks in the completions directory are absolute symlinks that resolve to a file in a plugin directory. These two commits ensure:
1) That the symlink resolves to a readable file
2) That the code extracting the plugin ID from a relative symlink to a non-plugin path does not get stuck in an infinite loop.

### DIFF
--- a/zinit-autoload.zsh
+++ b/zinit-autoload.zsh
@@ -501,7 +501,7 @@ ZINIT[EXTENDED_GLOB]=""
     # This in effect works as: "if different, then readlink"
     [[ -n "$tmp" ]] && in_plugin_path="$tmp"
 
-    if [[ "$in_plugin_path" != "$cpath" ]]; then
+    if [[ "$in_plugin_path" != "$cpath" && -r "$in_plugin_path" ]]; then
         # Get the user---plugin part of path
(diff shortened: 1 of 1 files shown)


alichtman (approved):

pschmitt:
When does 2. happen in?

alichtman:
Point 1 is definitely reasonable.

nbrown:
> 

I don't know how the symlink appeared, but I had a relative symlink to a non-plugin directory. The while loop processing that symlink's path assumes it will be an absolute path and be a subdirectory of ZINIT[PLUGIN_DIR]:

`while [[ "$in_plugin_path" != ${ZINIT[PLUGINS_DIR]}/[^/]## && "$in_plugin_path" != "/" ]]; do`

Adding the check against "." prevents the infinite loop.
`while [[ "$in_plugin_path" != ${ZINIT[PLUGINS_DIR]}/[^/]## && "$in_plugin_path" != "/" && "$in_plugin_path" != "." ]]; do`

vladdoster (approved):
LGTM!

PR #158, opened 2022-01-07
(bugfix): update bpick regex groups

vladdoster:
- correctly select `musl`
- correctly select `macOS`

A follow-up PR will add an update to select ARM  & `macOS`.


pschmitt:
We need to add tests for these.

AndrewDaws:
@vladdoster This really needs to be reverted or properly fixed, since this is breaking Linux program downloads as noted in issue #160. Something as commonly used as fzf is downloading the Darwin binary, rather than the Linux binary.

I was banging my head against a wall for hours trying to figure out why my setup was broken before I tried manually placing the Linux binary which lead me here.

Issue #160, opened 2022-01-10
🐞 bug: bick default selects darwin when using manjaro


gaerfield:
### Issue description

When "bpick" is not given explicitly the automatic selection of a release-binary selects darwin instead of linux.
```
❯ uname -a
Linux ithaka 5.13.19-2-MANJARO #1 SMP PREEMPT Sun Sep 19 21:31:53 UTC 2021 x86_64 GNU/Linux
```
For example [k9s](https://github.com/derailed/k9s/releases) is configured:
```
zcommand from"gh-r"; zload derailed/k9s
```

Installing / zinit update reports:
```
Downloading derailed/k9s…
(Requesting `k9s_Darwin_x86_64.tar.gz'…)
######################################################################## 100,0% 
ziextract: Unpacking the files from: `k9s_Darwin_x86_64.tar.gz'…
ziextract: Successfully extracted and assigned +x chmod to the file: `k9s'.
```

### zinit config

to much to be reported

### zinit version or commit ID

4da69625bcb7c12f929db84af2352e046eb11d75

### zsh version

5.8

### host info

XDG_SESSION_TYPE=x11 OSTYPE=linux-gnu CPUTYPE=x86_64 MACHTYPE=x86_64 5.13.19-2-MANJARO ID=manjaro

gaerfield:
Is it possible to set a fixed zinit version (to "tags/v3.17") to temporarily avoid zinit self-update updating to the latest unstable?

vladdoster:
See the zinit repository README.md for that option.

yutkat:
Now bpick on Linux is broken.
Please revert this commit https://github.com/zdharma-continuum/zinit/commit/4da69625bcb7c12f929db84af2352e046eb11d75

gaerfield:
@vladdoster Thanks for the hint. I was trying to find it before reporting and again now. Still I'm not able to identify the piece of information howto lock zinit to a specific version. I know that this is possible with "ver" for plugins, but for zinit itself I don't see this option.

gaerfield:
For the moment I have edited the function `.zinit-update-or-status-all` in `zinit-autoload.zsh` and commented line that executes a self-update when `zinit update` or `zinit update -p` is issued:
```
# .zinit-self-update -q
```
That prevents at least that zinit is updating itself and afterwards is overwriting my binaries with the darwin-variants during update.

AndrewDaws:
[...]

PR #161, opened 2022-01-11
Revert "(bugfix): update bpick regex groups"

AndrewDaws:
Reverts zdharma-continuum/zinit#158

As noted in issue #160 and pull request #158, this broke Linux application selection.


vladdoster:
Closing in favor of #162

gaerfield:
Why was this closed again?
Main is broken and self-update seems not to allow to pin a version. Reverting the main branch to be stable while improving the original feature and doing code-review and follow-up refinements without any pressure seems to be a valid approach.
Instead the main is broken for a week now.

alichtman (approved):

alichtman:
You're right -- we can't ever break the main branch.

vladdoster:
This goes back to breaking mac-os binary selections.

AndrewDaws:
> This goes back to breaking mac-os binary selections.

@vladdoster I investigated _all_ the commits made to the `zinit-install.zsh` file (I stopped after going back a year), and the only other changes made to that selection of code was your commit 2 months ago, b10e1c5fdcef3f77d6c46cbaf96e5c0b0ddc1a32, when you added `macos` as another search term for darwin. So for at least a year, that selection has _been_ broken for darwin in some way, so it's **NOT** a new development.

Should selection be fixed for darwin/mac-os?
YES! 100% it should be fixed. But it should **NOT**, nor ever be at the cost of breaking systems that had been working (nearly) perfectly for at least a year, if not more.

The moment this was reported as an issue, it should have been reverted so it could get back to a working state, then the changes could be worked on more until it was reviewed and shown to not break other things. The fact it was left in a broken state _AND_ this pull request closed without a fix or reverting is completely unacceptable.

I, and everyone else, are eternally grateful for the hard work and time people put in to open source projects like this, and certainly know I would not want it all to be dependent on me.

I just seriously worry about the health and longevity of this project if potentially breaking changes can be made and merged in by a single person without requiring either a completed review from someone else or at least some time passed to give others a chance to try and review. Let alone issues/pull requests addressing issues being closed for said issues.

Is there the possibility that someone still doesn't review a change?
Of course! That's an unfortunate downside of comparatively small projects like this, but at least people have a chance to make sure it doesn't break.

We don't need a repeat of psprint in which we are all at the fate of one person making rash decisions and changes that affect all users. I thought the whole point of making this more of a community-based project was to avoid exactly that issue.

AndrewDaws:
I should probably clarify that I do not want any witch hunts, fingers being pointed, or any drama. Since none of that ever helps.


I certainly know I break things at work and also fix things other people break. Bugs are a natural part of software development and engineering.

The important part though is that everyone acts professionally and things get fixed in some way shape or form (doesn't matter who!). Which fundamentally is all anyone actually cares about.

Avoiding bugs as much as possible, and fixing them when they inevitably they occur.

Issue #165, opened 2022-01-14
🎁 Feature request: support extraction of bz2 archives (*.tbz)


Gerrit-K:
### Feature request

It would be nice if `ziextract` also supports bz2 (or tbz) archives. 

For instance:
```bash
zi from'gh-r' for \
    sbin'btop' aristocratos/btop
```
results in:
```
Downloading aristocratos/btop…
(Requesting `btop-x86_64-macos-bigsur.tbz'…)
######################################################################## 100.0%
ziextract: WARNING: didn't recognize the archive type of `btop-x86_64-macos-bigsur.tbz'  }(no extraction has been done).%f%b
No files for compilation found.
Warning: ∞zinit-compile-plugin-hook hook returned with 1
```

vladdoster:
Functionality added in #166.

PR #162, opened 2022-01-12
(bugfix): `gh-r` choosing wrong binary for OS

vladdoster:
- fix #160
- fix zinit selecting Linux version on macOS
- fix zinit not selecting `musl` version on `musl` systems
- add `gh-r` ice unit tests on Ubuntu & macOS to verify behavior on
  different OSes

Signed-off-by: Vladislav Doster <mvdoster@gmail.com>

PR #166, opened 2022-01-14
Add support for `*.tbz` files

Gerrit-K:
Apparently bzipped archives are already supported. The file extension `tbz` is just not recognized.

Implements #165

### DIFF
--- a/zinit-install.zsh
+++ b/zinit-install.zsh
@@ -1571,7 +1571,7 @@ ziextract() {
         # First try known file extensions
         local -a files
         integer ret_val
-        files=( (#i)**/*.(zip|rar|7z|tgz|tbz2|tar.gz|tar.bz2|tar.7z|txz|tar.xz|gz|xz|tar|dmg|exe)~(*/*|.(_backup|git))/*(-.DN) )
+        files=( (#i)**/*.(zip|rar|7z|tgz|tbz|tbz2|tar.gz|tar.bz2|tar.7z|txz|tar.xz|gz|xz|tar|dmg|exe)~(*/*|.(_backup|git))/*(-.DN) )
         for file ( $files ) {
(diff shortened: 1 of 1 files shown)


zinit-install.zsh
```diff
@@ -1678,7 +1678,7 @@ ziextract() {
         ((#i)*.rar)
             →zinit-extract() { →zinit-check unrar "$file" || return 1; command unrar x "$file"; }
             ;;
-        ((#i)*.tar.bz2|(#i)*.tbz2)
+        ((#i)*.tar.bz2|(#i)*.tbz|(#i)*.tbz2)
```
vladdoster, line 1681: This can be reduced to avoid needing to duplicate the `tbz` but maybe not desired for readability.

LGTM otherwise, nice change.

vladdoster (approved):
+1 for fixing low hanging fruit

zinit-install.zsh
Gerrit-K: I must admit I'm drawing a blank here 😕 I've tried `tbz?`, `tbz?(2)` and even `tbz2?` (regex-style), but none of those variations worked. Do you mind helping me out here? Or do we want to keep it as it is?

pschmitt: Lgtm like it is. Readability is king in my eyes. We don't want zinit to get more cryptic ;)

pschmitt (approved):

zinit-install.zsh
Gerrit-K: Hehe, alright :)

Gerrit-K:
Regarding the failed doc check, I assume it's not caused by this PR and already targeted by https://github.com/zdharma-continuum/zinit/pull/162, right? Let me know if I can be of any help there :v:

Issue #137, opened 2021-12-13
🐞 bug: long git log in zinit update pauses update process


jerryc05:
### Issue description

when `zinit update` encounters a repo that has very long git logs, the update process will pause (because now the terminal is in `git log` mode) and will only proceed after I press `Q` to quit viewing git logs

### zinit config

_Nothing fancy and does not matter_

### zinit version or commit ID

Just updated, cannot remember previous zinit version

### zsh version

Just updated, cannot remember previous zinit version

### host info

HOSTTYPE=x86_64 OSTYPE=linux-gnu CPUTYPE=x86_64 MACHTYPE=x86_64 5.10.60.1-microsoft-standard-WSL2 ID=debian

pschmitt:
Thanks for reporting this. Forgot about that particular issue, but it's been grinding my gears for a few years :)

PR #177, opened 2022-02-06
Don't page git log for updates

nbrown:
When updating zinit or plugins, don't page git log outputs #137


pschmitt:
lgtm, thanks!

PR #184, opened 2022-02-20
*autoload: Make retval a local variable

NICHOLAS85:
8355489 Added added a retval variable which is a global variable when a non-parallelized `zinit update` was run. This simply localizes it for parallel and non-parallel runs.


vladdoster (approved):
LGTM!

PR #186, opened 2022-02-22
zinit zshelldoc update

alichtman:
We have a bunch of PRs that contain this large doc update, obscuring the
actual changes that are being made. This is a one-time, large-scale doc
update. Not sure how things got this out of sync, but we should not have
to do this again.

### DIFF
--- a/doc/zsdoc/zinit-autoload.zsh.adoc
+++ b/doc/zsdoc/zinit-autoload.zsh.adoc
@@ -25,8 +25,8 @@ FUNCTIONS
  .zinit-check-which-completions-are-installed
  .zinit-clear-completions
  .zinit-clear-report-for
- .zinit-compiled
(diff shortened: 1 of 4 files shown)


NICHOLAS85:
I'm not sure what's causing the issue, but it seems the output largely differs in order more than anything else (I haven't taken a deep dive). For example, running `make doc` on d1cb4e2b47cee6f298c59b35954f45aba9c26246 outputs the docs with a different order on my machine when it should've outputted the same due to no code changes. Could it be due to differing Zsh versions? Or different host OS.
I'm running zsh 5.8.1 on Fedora 35

pschmitt:
`zsd` is quick picky about the env it is run on. In my experience stuff like `LC_COLLATE` etc influence the output. Will look into dockerizing what's happening when running `make docs`

PR #187, opened 2022-02-22
[CI] Only run git diff once

alichtman:
We can return the exit code from the same command we print the diff
from.

### DIFF
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -46,12 +46,8 @@ jobs:
             /tmp/doc.tar.gz
 
       - name: 🕵️ Check if the zsd docs need to be updated
-      # NOTE: We need to run git diff twice here since w/o --quiet it's RC
-      # is always 0. The first is to display the diff output, the second one
-      # to test whether the diff is empty.
         run: |
-          git --no-pager diff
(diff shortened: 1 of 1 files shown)


.github/workflows/documentation.yaml
```diff
@@ -46,18 +46,14 @@ jobs:
...
-          git --no-pager diff
-          if git --no-pager diff --quiet
+          if git --no-pager diff --exit-code
           then
             echo "✅ The zshelldocs are up to date"
           else
             {
               echo "❌ The zshelldocs in the repo are not up to date"
               echo 'Please regenerate them with $ make doc'
-            } >&2
```
pschmitt, line 61: This wasn't intentional I assume.

alichtman: It was. It spits out the git diff output in all cases, allowing the omission of the second git diff command just to get the return code.

pschmitt: I meant this line. Why not show the error message on stderr?

pschmitt:
The test is failing because there *is* a diff. The issue here is with `make docs` ie. `zsd`, not git diff.

alichtman:
I know. This is a CI optimization PR, not a fix for the broken tests

Issue #170, opened 2022-01-29
🏦 Proposal: allow user to change tree to other program (e.g. exa)


Freed-Wu:
### Proposal

By default, `zinit ls` is

![Screenshot from 2022-01-29 16-26-57](https://user-images.githubusercontent.com/32936898/151653929-2eb58dab-42ce-4de7-a30d-e97d20136502.png)

If we change zinit-autoload.zsh:3247

```zsh
        list=( "${(f@)"$(LANG=en_US.utf-8 tree -L 3 --charset utf-8)"}" )
```

to

```zsh
        list=( "${(f@)"$(LANG=en_US.utf-8 tree -L 3 -C --charset utf-8)"}" )
```

it will be

![Screenshot from 2022-01-29 16-27-05](https://user-images.githubusercontent.com/32936898/151653980-28a5c234-d37c-4764-8c1a-4557e28b03dd.png)

colorful is good, I think.

However, if we allow user to use exa, that is

```zsh
        list=( "${(f@)"$(LANG=en_US.utf-8 exa --color=always --tree --icons -L3)"}" )
```

![Screenshot from 2022-01-29 16-29-04](https://user-images.githubusercontent.com/32936898/151654046-605be0cf-5743-4c71-bf76-6d383fbe7315.png)

we will see icons, I think it will be better.

Thanks.

PR #190, opened 2022-02-28
Second attempt to enable custom tree for 'zinit ls' command

nbrown:
Enable users to specify the zinit ls command via ZINIT[LIST_COMMAND] and
provide default commands for exa and tree. Prefer exa if it is installed.

Addresses issue https://github.com/zdharma-continuum/zinit/issues/170

I cherry-picked the original changes to a clean branch and then updated the docs to attempt to address the documentation action issues.

### DIFF
--- a/zinit-autoload.zsh
+++ b/zinit-autoload.zsh
@@ -3246,7 +3246,8 @@ EOF
         setopt localoptions extendedglob nokshglob noksharrays
         builtin cd -q "${ZINIT[SNIPPETS_DIR]}"
         local -a list
-        list=( "${(f@)"$(LANG=en_US.utf-8 tree -L 3 --charset utf-8)"}" )
+        local -x LANG=en_US.utf-8
+        list=( "${(f@)"$(${=ZINIT[LIST_COMMAND]})"}" )
         # Oh-My-Zsh single file
         list=( "${list[@]//(#b)(https--github.com--(ohmyzsh|robbyrussel)l--oh-my-zsh--raw--master(--)(#c0,1)(*))/$ZINIT[col-info]Oh-My-Zsh$ZINIT[col-error]${match[2]/--//}$ZINIT[col-pname]${match[3]//--/$ZINIT[col-error]/$ZINIT[col-pname]} $ZINIT[col-info](single-file)$ZINIT[col-rst] ${match[1]}}" )
         # Oh-My-Zsh SVN
--- a/doc/zsdoc/zinit-autoload.zsh.adoc
+++ b/doc/zsdoc/zinit-autoload.zsh.adoc
@@ -25,8 +25,8 @@ FUNCTIONS
  .zinit-check-which-completions-are-installed
  .zinit-clear-completions
  .zinit-clear-report-for
- .zinit-compiled
  .zinit-compile-uncompile-all
+ .zinit-compiled
  .zinit-confirm
  .zinit-create
  .zinit-delete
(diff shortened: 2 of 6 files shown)


vladdoster:
oof

alichtman:
Whatever. Let the test fail.

alichtman (approved):
Haven't tested it, but looks good to me

alichtman (approved):
Haven't tested it, but looks good to me

nbrown:
Well, I managed to make it work by using a docker container used in https://github.com/nektos/act

My conclusion is that there is something different between my macOS en_US.UTF-8 and the one used in the github actions containers.

alichtman:
This PR shouldn't contain a massive docs update. That should already be on master. I don't have time to jump into debugging it but I won't merge it with an ~800 line addition diff

vladdoster:
@nbrown great job on getting this to pass 

giphy

vladdoster:
@nbrown Since you mentioned this, I did some digging and think I found the root issue. 

I went to our Dockerfile and saw we don't explicitly set the locale, which seemed odd. A few GitHub issues later and I found this issue:

https://github.com/dotnet/runtime/issues/962

Your hypothesis seems to be spot on and created an issue to track it.

nbrown:
> This PR shouldn't contain a massive docs update. That should already be on master. I don't have time to jump into debugging it but I won't merge it with an ~800 line addition diff

Sorry for the large doc change, I don't know how you want to handle it but two choices are minimal doc changes that don't pass the documentation check or a large change that passes the check. I'll gladly revert it to not pass that check if that's what is preferred.

alichtman:
Yeah, let's keep the PR tight and drop the docs change. We'll get the docs test sorted out soon (maybe, idk -- no promises) but I'm fine with it failing here.

nbrown:
> Yeah, let's keep the PR tight and drop the docs change. We'll get the docs test sorted out soon (maybe, idk -- no promises) but I'm fine with it failing here.

Done. I reverted the doc fix so it's now a minimal change.

On the docs test, I'll leave some info on that bug. I think it primarily comes down to a sorting (collation) inconsistency between Linux and macOS/BSD based on my reading and testing.

zinit-autoload.zsh
```diff
@@ -3246,7 +3246,8 @@ EOF
         setopt localoptions extendedglob nokshglob noksharrays
         builtin cd -q "${ZINIT[SNIPPETS_DIR]}"
         local -a list
-        list=( "${(f@)"$(LANG=en_US.utf-8 tree -L 3 --charset utf-8)"}" )
+        local -x LANG=en_US.utf-8
```
alichtman, line 3249: Is this required by `exa`?

nbrown: Actually no. I took the issue's suggestion at face value, but as far as I can tell from exa's source code, it doesn't rely on locale for sorting.

vladdoster (commented):
This check evaluates to true even though I have exa, but no tree.

zinit-autoload.zsh
vladdoster, line 3249: The [prior check](https://github.com/zdharma-continuum/zinit/pull/190/files#diff-6a3752f1e40118bbfc5b0e05b9fa86aaffbd897b4c36f760bb3221a862f68343R3240-R3244) evaluates to false even though I have exa, but no tree.

vladdoster (changes requested):

vladdoster (approved):

vladdoster:
Closing in favor of #221 due to lack of update on the PR.

PR #196, opened 2022-03-12
Enable developers to build docs in docker

nbrown:
After my struggles to generate the docs such that they pass the documentation check, I used docker to execute ``make doc``.

I also added ``doc-in-docker`` as a target in the Makefile which will build the docker image and generate the docs in the current git clone.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,4 @@ zmodules/Src/zdharma/curses_keys.h
 zmodules/Test/*.tmp
 /.project
 doc/zsdoc/data
+doc/docker/image-id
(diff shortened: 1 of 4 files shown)


pschmitt:
Interesting work, though I wonder why not just use the zinit docker image - instead of building a separate one just for generating docs?

https://github.com/zdharma-continuum/zinit/pkgs/container/zinit

nbrown:
> Interesting work, though I wonder why not just use the zinit docker image - instead of building a separate one just for generating docs?
> 
> https://github.com/zdharma-continuum/zinit/pkgs/container/zinit

I tried to use an alpine-based docker, but the musl-based locale I found did not generate the same docs as the documentation github action does. I found that I could match it with a ubuntu-based image.

pschmitt:
The best solution to this problem would IMO be to use the docker container for building the docs in both envs: locally, and in CI.
No more guesswork then.

vladdoster:
Closing in favor of #200

PR #185, opened 2022-02-22
[DNM] testing: zshelldocs generation bug

vladdoster:
Signed-off-by: Vladislav Doster <mvdoster@gmail.com>

PR #200, opened 2022-03-17
Use container for generating docs

pschmitt:
So here we go. It's all using the zinit container.

Refs:
- #185
- #196

### DIFF
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -21,9 +21,7 @@ jobs:
       - name: 👷 Generate zsd
         run: |
           make doc-container
-          rc="$?"
           sudo chown -R "$(id -u):$(id -g)" .
-          exit "$rc"
 
       - name: 📇 Generate PDF
         run: |
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -5,14 +5,6 @@ jobs:
   zshelldoc:
     runs-on: ubuntu-latest
     steps:
-      - name: 📦 Install zshelldoc
-        run: |
-          sudo apt-get install -y zsh
-          git clone https://github.com/zdharma-continuum/zshelldoc.git /tmp/zsd.git
-          cd /tmp/zsd.git
-          sudo make install
-          zsd --help
-
       - name: ♦️ Setup ruby
         uses: ruby/setup-ruby@v1
         with:
@@ -28,7 +20,10 @@ jobs:
 
       - name: 👷 Generate zsd
         run: |
-          make doc
+          make doc-container
+          rc="$?"
+          sudo chown -R "$(id -u):$(id -g)" .
+          exit "$rc"
 
       - name: 📇 Generate PDF
         run: |
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,9 @@ testD:
 testE:
 	make VERBOSE=$(VERBOSE) NODIFF=$(NODIFF) DEBUG=$(DEBUG) OPTDUMP=$(OPTDUMP) OPTS="ignoreclosebraces" -C test test
 
+doc-container: zinit.zsh zinit-side.zsh zinit-install.zsh zinit-autoload.zsh
+	./scripts/docker-run.sh --docs --debug
+
 # Set LC_ALL to avoid having to deal with different locales.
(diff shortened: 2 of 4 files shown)


.github/workflows/documentation.yaml
```diff
@@ -28,7 +20,10 @@ jobs:
 
       - name: 👷 Generate zsd
         run: |
-          make doc
+          make doc-container
+          rc="$?"
+          sudo chown -R "$(id -u):$(id -g)" .
```
vladdoster, line 25: Why is this needed?

scripts/docker-run.sh
```diff
@@ -53,6 +55,16 @@ run() {
 
   local -a args=(--rm)
 
+  local cruntime=docker
+  local sudo_cmd
+  if [[ -z "$CI" ]] && command -v podman >/dev/null 2>&1
+  then
+    cruntime=podman
+    # rootless containers are a PITA
+    # https://www.tutorialworks.com/podman-rootless-volumes/
```
vladdoster, line 64: Is this really needed? Do we test both runtimes? I think it adds complexity while adding no real benefit.

1. Why not use [docker w/ rootless](https://docs.docker.com/engine/security/rootless/
)?
1. Do we want to support singularity containers, too?

Makefile
```diff
@@ -20,6 +20,9 @@ testD:
 testE:
 	make VERBOSE=$(VERBOSE) NODIFF=$(NODIFF) DEBUG=$(DEBUG) OPTDUMP=$(OPTDUMP) OPTS="ignoreclosebraces" -C test test
 
+doc-container: zinit.zsh zinit-side.zsh zinit-install.zsh zinit-autoload.zsh
```
vladdoster, line 23: Can we extract: 

```
zinit.zsh zinit-side.zsh zinit-install.zsh zinit-autoload.zsh
```

into variable to avoid duplicating it in `doc` target?

.github/workflows/documentation.yaml
```diff
@@ -28,7 +20,10 @@ jobs:
 
       - name: 👷 Generate zsd
         run: |
-          make doc
+          make doc-container
+          rc="$?"
```
vladdoster, line 24: Won't GH actions exit on their own if the `1` exit code is returned? This seems redundant

.github/workflows/documentation.yaml
pschmitt: There's an explicit chown to the user running in the container in the command run by the container. This is resetting the perms afterwards, w/o it the PDF step fails for example.

scripts/docker-run.sh
pschmitt: What's your point? "Please remove this feature"?

The cli of both docker and podman are almost 100% the same. Besides using a different command name there is pretty much no added complexity here.

I want to support both, cause I don't want contributors to go through the hassle of installing podman or docker even though their distro ships the other.

Makefile
pschmitt: Maybe? I don't know how :D

.github/workflows/documentation.yaml
pschmitt: Right, it does. Will remove this.

scripts/docker-run.sh
vladdoster: Your comment on line 63 made me ask the question

> # rootless containers are a PITA

Additionally, Podman presently only supports running containers on Linux. However, we are building a remote client which can run on Windows and macOS and manage Podman containers on a Linux system via the REST API using SSH tunneling.

Feels like a solution to a problem no one has encountered (e.g., pre-optimization).

Makefile
vladdoster: No worries, just a nit :^)

scripts/docker-run.sh
pschmitt: Well, well, well. This all started cause I didn't want to:

1. install docker 
2. Argue with people about container runtimes

Sigh.

.github/workflows/documentation.yaml
pschmitt: Done.

[78c898e](https://github.com/zdharma-continuum/zinit/pull/200/commits/78c898ea775191714c22155522da6054aa034a3e)

alichtman (approved):

Issue #199, opened 2022-03-14
[bug]: atclone/mv/... are not run when installing plugins non-interactively


jankatins:
### Issue description

When I remove my whole plugin dir, the direnv hock is not run, resulting in an error. If I remove only the direnv directory, everything is fine:

```bash
λ  rm -rf .zinit/plugins/

# Opening a new tab
... lots of zinit install ...
Downloading direnv/direnv…
(Requesting `direnv.linux-amd64'…)
############################################################################################################################################################################################################ 100.0%
ziextract: Successfully extracted and assigned +x chmod to the file: `direnv.linux-amd64'.
bin-gem-node annex: Warning: The sbin'' ice (`direnv') didn't match any files
.zinit-load-plugin:source:110: no such file or directory: /home/jan/.zinit/plugins/direnv---direnv/zhook.zsh
```

When I just remove the direnv, it works:

```bash
λ  rm -rf .zinit/plugins/direnv---direnv

Downloading direnv/direnv…
(Requesting `direnv.linux-amd64'…)
############################################################################################################################################################################################################ 100.0%
ziextract: Successfully extracted and assigned +x chmod to the file: `direnv.linux-amd64'.
No files for compilation found.
Warning: ∞zinit-compile-plugin-hook hook returned with 1
bin-gem-node annex: Created the direnv shim and set +x on the direnv binary
```

I tried several way to work around this (e.g. with or without `mv"direnv* -> direnv"` or `sbin"direnv* -> direnv"`, with dot in the mv or without), but none of them worked. Even running the direnv line directly after installing the sbin plugin had the same effect: if I remove th ewhol plugin dir, it fails to mv the direnv binary to `direnv` and does not run the atclone line.

### zinit config

```zsh
  source ~/.zinit/bin/zinit.zsh

  zinit light-mode for zdharma-continuum/zinit-annex-bin-gem-node

  # some oh-my-zsh niceties
  # OMZ plugins: https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins
  zinit wait lucid for \
    atinit"ZINIT[COMPINIT_OPTS]=-C; zicompinit; zicdreplay"  \
          OMZL::key-bindings.zsh \
          OMZL::history.zsh \
          OMZP::command-not-found \
          OMZP::extract \
          OMZP::colored-man-pages \
          OMZL::completion.zsh \
          OMZL::termsupport.zsh \
          zdharma-continuum/fast-syntax-highlighting \
[...]

PR #227, opened 2022-04-25
fix: allow zinit to be run from non-interactive scripts

jankatins:
Exclamation marks (per default) get escaped in interactive mode, but not in non-interactive modes. In this case it meant that zinit would not find any hooks which implements the ices when running in non-interactive mode (e.g. in a script). This in turn would NOT result in errors but would just silently not execute any of the ices which of course would be a bad thing(tm).

This actually happened in the zunit tests as they run non-interactively and therefore any ices based on exclamation mark hooks were not ran. I fixed the few places where (after the fix in this PR) the ice were actually ran and errored due to mis-specifications. I also added explicit checks for a few ices. -> before this PR, ices are basically not tested (and the exclamation mark hook based ones not testable at all!) :-(

## Description

This a) disables history mark expansion during hook register and changes any search pattern to not include the (double) escaped exclamation mark 

## Motivation and Context

It broke using zinit calls with ices in scripts (e.g. from a $HOME bootstrap script or an update script) 

https://github.com/zdharma-continuum/zinit/issues/199 has the details...

Closes: #199

## How Has This Been Tested?

See the new zunit tests and a few fixes in old tests which were needed because the actual ices were not run beforehand and therefore didn't error when they were not specified correctly.

Running the same ice (mv in this case)  interactively and in a script. Both calls contain the `direnv.darwin-amd64 -> direnv` log line.

```zsh

[16:30:23] λ  zinit delete direnv/direnv -y ; zsh -i -c -f --no-rcs 'source ~/.zinit/bin/zinit.zsh ; source ~/.zinit/bin/zinit-install.zsh; zinit from"gh-r" mv"direnv* -> direnv" for direnv/direnv'
bin-gem-node annex: The direnv shim didn't exist in $ZPFX/bin (or isn't a regular file)

Done (action executed, exit code: 0)

Downloading direnv/direnv…
(Requesting `direnv.darwin-amd64'…)
#################################################################################################################################################################################################################### 100.0%
ziextract: Successfully extracted and assigned +x chmod to the file: `direnv.darwin-amd64'.
direnv.darwin-amd64 -> direnv
No files for compilation found.
Warning: ∞zinit-compile-plugin-hook hook returned with 1


[16:30:29] [1] ✗  ls -la ~/.zinit/plugins/direnv---direnv/direnv*
-rwxr-xr-x  1 jankatins  staff  8811104 Apr 25 16:30 /Users/jankatins/.zinit/plugins/direnv---direnv/direnv


[16:30:37] λ  zinit delete direnv/direnv -y ; source ~/.zinit/bin/zinit.zsh ; source ~/.zinit/bin/zinit-install.zsh; zinit from"gh-r" mv"direnv* -> direnv" for direnv/direnv
bin-gem-node annex: The direnv shim didn't exist in $ZPFX/bin (or isn't a regular file)

Done (action executed, exit code: 0)

Downloading direnv/direnv…
(Requesting `direnv.darwin-amd64'…)
#################################################################################################################################################################################################################### 100.0%
ziextract: Successfully extracted and assigned +x chmod to the file: `direnv.darwin-amd64'.
direnv.darwin-amd64 -> direnv
No files for compilation found.
Warning: ∞zinit-compile-plugin-hook hook returned with 1

[16:31:02] [1] ✗  ls -la ~/.zinit/plugins/direnv---direnv/direnv*
-rwxr-xr-x  1 jankatins  staff  8811104 Apr 25 16:31 /Users/jankatins/.zinit/plugins/direnv---direnv/direnv
```

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -72,6 +72,9 @@ jobs:
       - name: "gh-r"
         run: zunit run tests/gh-r.zunit
 
+      - name: "ices"
+        run: zunit run tests/ices.zunit
+
       - name: "plugins"
         run: zunit run tests/plugins.zunit
 
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ txt/
 zmodules/
 tests/_output/
 tests/_support/zunit/
+.idea
--- a/tests/gh-r.zunit
+++ b/tests/gh-r.zunit
@@ -145,7 +145,7 @@
   $dive --version; assert $state equals 0
 }
 @test 'docker-buildx' { # A monitor of resources
-  run zinit for as'completions' atclone'buildx* completion zsh > _buildx' lbin'!buildx-* -> buildx' @docker/buildx
+  run zinit for as'completions' atclone'./buildx* completion zsh > _buildx' lbin'!buildx-* -> buildx' @docker/buildx
   assert $state equals 0
   local buildx="$ZBIN/buildx"; assert "$buildx" is_executable
   $buildx version; assert $state equals 0
--- a/tests/ices.zunit
+++ b/tests/ices.zunit
@@ -1,0 +1,52 @@
+#!/usr/bin/env zunit
+
+@setup {
+  ZPLUGINS=$ZINIT[PLUGINS_DIR]
+
+  function _zunit_assert_not_exists() {
+    local pathname=$1 filepath
+
+    # If filepath is relative, prepend the test directory
+    if [[ "${pathname:0:1}" != "/" ]]; then
+      filepath="$testdir/${pathname}"
+    else
+      filepath="$pathname"
+    fi
+
+    [[ ! -e "$filepath" ]] && return 0
+
+    echo "'$pathname' does exist"
+    exit 1
+  }
+
+}
+
+@test 'mv' {
+  run zinit as"null" id-as"test/mv" mv"readme.md -> mv.md" for zdharma-continuum/null
+  assert $state equals 0
+  assert "$ZPLUGINS/test---mv/mv.md" is_file
+  assert "$ZPLUGINS/test---mv/mv.md" is_file
+  assert "$ZPLUGINS/test---mv/readme.md" not_exists
+}
+
+@test 'cp' {
+  run zinit as"null" id-as"test/cp" cp"readme.md -> cp.md" for zdharma-continuum/null
+  assert $state equals 0
+  assert "$ZPLUGINS/test---cp/cp.md" is_file
+  assert "$ZPLUGINS/test---cp/cp.md" is_readable
+  assert "$ZPLUGINS/test---cp/readme.md" is_file
+}
+
+@test 'atclone' {
+  run zinit as"null" id-as"test/atclone" atclone"mv readme.md atclone.md" for zdharma-continuum/null
+  assert $state equals 0
+  assert "$ZPLUGINS/test---atclone/atclone.md" is_file
+  assert "$ZPLUGINS/test---atclone/atclone.md" is_readable
+  assert "$ZPLUGINS/test---atclone/readme.md" not_exists
+}
+
+@test 'make' {
+  run zinit as"null" id-as"test/make" atclone"printf 'all:\n\ttouch whatever\n' > Makefile" make"" for zdharma-continuum/null
+  assert $state equals 0
+  assert "$ZPLUGINS/test---make/whatever" is_file
+}
--- a/tests/snippets.zunit
+++ b/tests/snippets.zunit
@@ -52,9 +52,12 @@
 }
 @test 'tldr-completion::gh-r' {
   artifact="$ZINIT[PLUGINS_DIR]/tldr-completion---gh-r"
-  run zinit for as"completion" from"gh-r" id-as'tldr-completion/gh-r' bpick"completions_zsh" @dbrgn/tealdeer
+  run zinit for as"completion" from"gh-r" id-as'tldr-completion/gh-r' bpick"completions_zsh" mv"completions_zsh -> _tldr_ghr" pick"_tldr_ghr" @dbrgn/tealdeer
   assert $state equals 0
-  assert "$artifact/completions_zsh" is_file; assert $artifact is_readable
+  assert "$artifact/_tldr_ghr" is_file
+  assert $artifact/_tldr_ghr is_readable
+  assert "$ZINIT[COMPLETIONS_DIR]"/_tldr_ghr is_file
+  assert "$ZINIT[COMPLETIONS_DIR]"/_tldr_ghr is_readable
 }
 @test 'zinit local snippet symlink' {
   mkdir -p "${ZINIT[HOME_DIR]}/external"
--- a/zinit-install.zsh
+++ b/zinit-install.zsh
@@ -2360,18 +2360,23 @@ zimv() {
 
     @zinit-substitute from to
 
-    local -a afr
+    local -a afr retval
     ( () { setopt localoptions noautopushd; builtin cd -q "$dir"; } || return 1
       afr=( ${~from}(DN) )
       if (( ${#afr} )) {
           if (( !OPTS[opt_-q,--quiet] )) {
               command cp -vf "${afr[1]}" "$to"
+              retval=$?
+              # ignore errors if no compiled file is found
               command cp -vf "${afr[1]}".zwc "$to".zwc 2>/dev/null
           } else {
               command cp -f "${afr[1]}" "$to"
+              retval=$?
+              # ignore errors if no compiled file is found
               command cp -f "${afr[1]}".zwc "$to".zwc 2>/dev/null
           }
       }
+      return $retval
     )
 }
 # ]]]
--- a/zinit-install.zsh
+++ b/zinit-install.zsh
@@ -451,9 +451,9 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
             # Store ices at clone of a plugin
             .zinit-store-ices "$local_path/._zinit" ICE "" "" "" ""
             reply=(
-                ${(on)ZINIT_EXTS2[(I)zinit hook:\\\!atclone-pre <->]}
-                ${(on)ZINIT_EXTS[(I)z-annex hook:\\\!atclone-<-> <->]}
-                ${(on)ZINIT_EXTS2[(I)zinit hook:\\\!atclone-post <->]}
+                ${(on)ZINIT_EXTS2[(I)zinit hook:\!atclone-pre <->]}
+                ${(on)ZINIT_EXTS[(I)z-annex hook:\!atclone-<-> <->]}
+                ${(on)ZINIT_EXTS2[(I)zinit hook:\!atclone-post <->]}
             )
             for key in "${reply[@]}"; do
                 arr=( "${(Q)${(z@)ZINIT_EXTS[$key]:-$ZINIT_EXTS2[$key]}[@]}" )
@@ -999,9 +999,9 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
                         # Run annexes' atpull hooks (the before atpull-ice ones).
                         # The SVN block.
                         reply=(
-                            ${(on)ZINIT_EXTS2[(I)zinit hook:e-\\\!atpull-pre <->]}
-                            ${${(M)ICE[atpull]#\!}:+${(on)ZINIT_EXTS[(I)z-annex hook:\\\!atpull-<-> <->]}}
-                            ${(on)ZINIT_EXTS2[(I)zinit hook:e-\\\!atpull-post <->]}
+                            ${(on)ZINIT_EXTS2[(I)zinit hook:e-\!atpull-pre <->]}
+                            ${${(M)ICE[atpull]#\!}:+${(on)ZINIT_EXTS[(I)z-annex hook:\!atpull-<-> <->]}}
+                            ${(on)ZINIT_EXTS2[(I)zinit hook:e-\!atpull-post <->]}
                         )
                         for key in "${reply[@]}"; do
                             arr=( "${(Q)${(z@)ZINIT_EXTS[$key]:-$ZINIT_EXTS2[$key]}[@]}" )
@@ -1093,9 +1093,9 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
                     # The URL-snippet block.
                     if [[ $update = -u && $ZINIT[annex-multi-flag:pull-active] -ge 1 ]] {
                         reply=(
-                            ${(on)ZINIT_EXTS2[(I)zinit hook:e-\\\!atpull-pre <->]}
-                            ${${ICE[atpull]#\!}:+${(on)ZINIT_EXTS[(I)z-annex hook:\\\!atpull-<-> <->]}}
-                            ${(on)ZINIT_EXTS2[(I)zinit hook:e-\\\!atpull-post <->]}
+                            ${(on)ZINIT_EXTS2[(I)zinit hook:e-\!atpull-pre <->]}
+                            ${${ICE[atpull]#\!}:+${(on)ZINIT_EXTS[(I)z-annex hook:\!atpull-<-> <->]}}
+                            ${(on)ZINIT_EXTS2[(I)zinit hook:e-\!atpull-post <->]}
                         )
                         for key in "${reply[@]}"; do
                             arr=( "${(Q)${(z@)ZINIT_EXTS[$key]:-$ZINIT_EXTS2[$key]}[@]}" )
@@ -1153,9 +1153,9 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
             # The local-file snippets block.
             if [[ $update = -u ]] {
                 reply=(
-                    ${(on)ZINIT_EXTS2[(I)zinit hook:e-\\\!atpull-pre <->]}
-                    ${${(M)ICE[atpull]#\!}:+${(on)ZINIT_EXTS[(I)z-annex hook:\\\!atpull-<-> <->]}}
-                    ${(on)ZINIT_EXTS2[(I)zinit hook:e-\\\!atpull-post <->]}
+                    ${(on)ZINIT_EXTS2[(I)zinit hook:e-\!atpull-pre <->]}
+                    ${${(M)ICE[atpull]#\!}:+${(on)ZINIT_EXTS[(I)z-annex hook:\!atpull-<-> <->]}}
+                    ${(on)ZINIT_EXTS2[(I)zinit hook:e-\!atpull-post <->]}
                 )
                 for key in "${reply[@]}"; do
                     arr=( "${(Q)${(z@)ZINIT_EXTS[$key]:-$ZINIT_EXTS2[$key]}[@]}" )
@@ -1248,9 +1248,9 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
             # Run annexes' atpull hooks (the before atpull-ice ones).
             # The block is common to all 3 snippet types.
             reply=(
-                ${(on)ZINIT_EXTS2[(I)zinit hook:no-e-\\\!atpull-pre <->]}
-                ${${ICE[atpull]:#\!*}:+${(on)ZINIT_EXTS[(I)z-annex hook:\\\!atpull-<-> <->]}}
-                ${(on)ZINIT_EXTS2[(I)zinit hook:no-e-\\\!atpull-post <->]}
+                ${(on)ZINIT_EXTS2[(I)zinit hook:no-e-\!atpull-pre <->]}
+                ${${ICE[atpull]:#\!*}:+${(on)ZINIT_EXTS[(I)z-annex hook:\!atpull-<-> <->]}}
+                ${(on)ZINIT_EXTS2[(I)zinit hook:no-e-\!atpull-post <->]}
             )
             for key in "${reply[@]}"; do
                 arr=( "${(Q)${(z@)ZINIT_EXTS[$key]:-$ZINIT_EXTS2[$key]}[@]}" )
@@ -1265,9 +1265,9 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
             # Run annexes' atclone hooks (the before atclone-ice ones)
             # The block is common to all 3 snippet types.
             reply=(
-                ${(on)ZINIT_EXTS2[(I)zinit hook:\\\!atclone-pre <->]}
-                ${(on)ZINIT_EXTS[(I)z-annex hook:\\\!atclone-<-> <->]}
-                ${(on)ZINIT_EXTS2[(I)zinit hook:\\\!atclone-post <->]}
+                ${(on)ZINIT_EXTS2[(I)zinit hook:\!atclone-pre <->]}
+                ${(on)ZINIT_EXTS[(I)z-annex hook:\!atclone-<-> <->]}
+                ${(on)ZINIT_EXTS2[(I)zinit hook:\!atclone-post <->]}
             )
             for key in "${reply[@]}"; do
                 arr=( "${(Q)${(z@)ZINIT_EXTS[$key]:-$ZINIT_EXTS2[$key]}[@]}" )
--- a/zinit.zsh
+++ b/zinit.zsh
@@ -1184,6 +1184,12 @@ builtin setopt noaliases
 # FUNCTION: @zinit-register-hook. [[[
 # Registers the z-annex inside Zinit – i.e. an Zinit extension
 @zinit-register-hook() {
+    builtin emulate -LR zsh ${=${options[xtrace]:#off}:+-o xtrace}
+    # zsh per default behaves differently in interactive sessions (escapes !) and
+    # non-interactive ones (does not). Setting 'nobanghist' option disables backslash
+    # escaping for "!" even in interactive sessions.
+    # Exclamation marks are used in a lot of hock names.
+    builtin setopt extendedglob nobanghist noshortloops typesetsilent warncreateglobal
     local name="$1" type="$2" handler="$3" icemods="$4" key="zinit ${(q)2}"
     ZINIT_EXTS2[seqno]=$(( ${ZINIT_EXTS2[seqno]:-0} + 1 ))
     ZINIT_EXTS2[$key${${(M)type#hook:}:+ ${ZINIT_EXTS2[seqno]}}]="${ZINIT_EXTS2[seqno]} z-annex-data: ${(q)name} ${(q)type} ${(q)handler} '' ${(q)icemods}"
@@ -1391,7 +1397,7 @@ builtin setopt noaliases
 
     ZINIT[CUR_USPL2]="$id_as" ZINIT_REPORTS[$id_as]=
 
-    reply=( ${(on)ZINIT_EXTS[(I)z-annex hook:\\\!atinit-<-> <->]} )
+    reply=( ${(on)ZINIT_EXTS[(I)z-annex hook:\!atinit-<-> <->]} )
     for key in "${reply[@]}"; do
         arr=( "${(Q)${(z@)ZINIT_EXTS[$key]}[@]}" )
         "${arr[5]}" snippet "$save_url" "$id_as" "$local_dir/$dirname" \!atinit || \
@@ -1453,7 +1459,7 @@ builtin setopt noaliases
         [[ -n ${ICE[multisrc]} ]] && { local ___oldcd="$PWD"; () { setopt localoptions noautopushd; builtin cd -q "$local_dir/$dirname"; }; eval "reply=(${ICE[multisrc]})"; () { setopt localoptions noautopushd; builtin cd -q "$___oldcd"; }; local fname; for fname in "${reply[@]}"; do ZERO="${${(M)fname:#/*}:-$local_dir/$dirname/$fname}"; (( ${+ICE[silent]} )) && { { [[ -n $precm ]] && { builtin ${precm[@]} 'source "$ZERO"'; ((1)); } || { ((1)); builtin source "$ZERO"; }; } 2>/dev/null 1>&2; (( retval += $? )); ((1)); } || { ((1)); { [[ -n $precm ]] && { builtin ${precm[@]} 'source "$ZERO"'; ((1)); } || { ((1)); builtin source "$ZERO"; }; }; (( retval += $? )); }; done; }
 
         # Run the atload hooks right before atload ice.
-        reply=( ${(on)ZINIT_EXTS[(I)z-annex hook:\\\!atload-<-> <->]} )
+        reply=( ${(on)ZINIT_EXTS[(I)z-annex hook:\!atload-<-> <->]} )
         for key in "${reply[@]}"; do
             arr=( "${(Q)${(z@)ZINIT_EXTS[$key]}[@]}" )
             "${arr[5]}" snippet "$save_url" "$id_as" "$local_dir/$dirname" \!atload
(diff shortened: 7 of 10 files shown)


jankatins:
There are still two places where `\\\!` exists: in `test-excl-id-as` and in zinit-autoload.zsh: https://github.com/zdharma-continuum/zinit/blob/main/zinit-autoload.zsh#L3324:

```zsh
    [[ -e $local_dir ]] && {
        for el ( ${ice_order[@]} ) {
            val="${ice[$el]}"
            cand1="${(qqq)val}"
            cand2="${(qq)val}"
            if [[ -n "$val" ]] {
                [[ "${cand1/\\\$/}" != "$cand1" || "${cand1/\\\!/}" != "$cand1" ]] && output+=( "$el$cand2" ) || output+=( "$el$cand1" )
            } elif [[ ${+ice[$el]} = 1 && -n "${nval_ices[(r)$el]}" ]] {
                output+=( "$el" )
            }
        }
```

In both cases, I do not know what to do with it :-(

Also: not sure if this should get a test (and how: given that tests would run in scripts and so would always be non-interactive?).

vladdoster:
For the failing documentation test, you should be able to run:

```bash
zinit make"PREFIX=$ZPFX install" for @zdharma-continuum/zshelldoc
make doc
```

vladdoster:
And if the unit test tests are failing after #226 is merged, I'll look and make sure it isn't related to my work over the weekend.

vladdoster:
this XKCD seems warranted ;)

jankatins:
Manually added the doc change, as I got a real bunch of changes here (mac? Different version?). Also fixed the commit message to be the same as the PR title

jankatins:
@vladdoster  Rebased and force pushed on top of the 386 fixes in #226

vladdoster:
Looks like it fails here

https://github.com/zdharma-continuum/zinit/blob/main/tests/gh-r.zunit#L260

jankatins:
A bit more tracing and this turned out to be these two lines:

```
             command cp -vf "${afr[1]}" "$to"
             command cp -vf "${afr[1]}".zwc "$to".zwc 2>/dev/null
```

It seems the return value of the second copy (which didn't find files to copy) was returned from the hook and therefore the hook run marked as failed. 

Verified with 

```
zsh -c 'source ~/.zshrc ; zinit delete github/hub -y; zinit cp"hub-*/etc/hub.zsh_completion -> _hub" sbin"hub-*/bin/hub" from"gh-r" as"null" for @github/hub '
```

@vladdoster See the second commit here for a possible fix (modeled after the pattern found in the mv hook)

jankatins:
BTW: is there any better way to enable tracing everywhere than `s/setopt / setopt xtrace/` in all files? Seems every time `emulate -LR zsh` is called in a function, xtrace is off again :-( E.g. I started testing with `zsh -c 'source ~/.zshrc ; export PS4="%? // %x.%I: "; exec 3>&2 2>/tmp/zinit.log ; setopt xtrace ; zinit delete github/hub -y; zinit cp"hub-*/etc/hub.zsh_completion -> _hub" sbin"hub-*/bin/hub" from"gh-r" as"null" for @github/hub ; unsetopt xtrace ; exec 2>&3 3>&-'` and missed most of the interesting trace until I ran the above search&replace...

jankatins:
Current failure:

```
✘ tldr-completion::gh-r
  
Downloading dbrgn/tealdeer… (at label: tldr-completion/gh-r…)

(Requesting `completions_zsh'…)



#=#=#                                                                         



######################################################################## 100.0%
ziextract: WARNING: didn't recognize the archive type of `completions_zsh'  }(no extraction has been done).%f%b

No files for compilation found.

Warning: ∞zinit-compile-plugin-hook hook returned with 1
'1' is not equal to '0'
```

I can reproduce locally and the fix is to add a `pick"completions_zsh"` to the test...


```zsh
λ  zinit delete "tldr-completion/gh-r" -y ; zinit for as"completion" from"gh-r" id-as"tldr-completion/gh-r" bpick"completions_zsh" pick"completions_zsh"  @dbrgn/tealdeer

Done (action executed, exit code: 0)

Downloading dbrgn/tealdeer… (at label: tldr-completion/gh-r…)
(Requesting `completions_zsh'…)
############################################################################################################################################################################################################ 100.0%
ziextract: WARNING: didn't recognize the archive type of `completions_zsh'  }(no extraction has been done).%f%b
Note: Compiling: completions_zsh… OK.
```

-> Pushed in a separate commit as I'm not completely sure if this need is expected and why this is now failing (no exclamation mark in the hook name) after the fix.

jankatins:
@vladdoster It's green :-)

vladdoster:
Even though its green, the workaround seems like there is a regression somewhere, no?

jankatins:
It feels more like the unit tests were having the same problem (I guess they are running non-interactive) and therefore config errors were not noticed. Eg nothing in the tests check for compiling in the tldr gh-r case or for the copied _hub completion file and so the missing hooks didn't least to errors. And running it interactively will get you a error message but the main functionality is there (hub command and completion for tldr).

tests/snippets.zunit
```diff
@@ -52,7 +52,7 @@
 }
 @test 'tldr-completion::gh-r' {
   artifact="$ZINIT[PLUGINS_DIR]/tldr-completion---gh-r"
-  run zinit for as"completion" from"gh-r" id-as'tldr-completion/gh-r' bpick"completions_zsh" @dbrgn/tealdeer
+  run zinit for as"completion" from"gh-r" id-as'tldr-completion/gh-r' bpick"completions_zsh" pick"completions_zsh" @dbrgn/tealdeer
```
jankatins, line 55: My understanding of the need for this is that the compile now fails without this because the compile step wasn't run before in the unit test because it was ran non-interactive and therefore triggering the bug which got fixed in this PR.

jankatins:
@vladdoster Rebased on top of latest main and still green :-)

I tried to add reasoning as inline comments why each add-on fix was needed. Happy to clarify if there is still anything unclear.

tests/snippets.zunit
jankatins: On main and my branch, this one works:

```zsh

[14:31:37] λ  zinit for as"completion" from"gh-r" id-as"tldr-completion/gh-r" bpick"completions_zsh" mv"completions_zsh -> _tldr" pick"_tldr" @dbrgn/tealdeer || echo bad

Downloading dbrgn/tealdeer… (at label: tldr-completion/gh-r…)
(Requesting `completions_zsh'…)
#################################################################################################################################################################################################################### 100.0%
ziextract: WARNING: didn't recognize the archive type of `completions_zsh'  }(no extraction has been done).%f%b
completions_zsh -> _tldr
Note: Compiling: _tldr… OK.
Installed 0 completions. They are stored in the $INSTALLED_COMPS array.
Skipped installing 1 completions. They are stored in the $SKIPPED_COMPS array.
```

The SKIPPED is because the symlink in ~/.zinit/completion/ seems to not get removed on delete:

```
[14:32:48] λ  zinit delete tldr-completion/gh-r -y

[14:33:56] λ  ls -la ../completions/_tldr
lrwxr-xr-x  1 jankatins  staff  60 May  6 14:30 ../completions/_tldr -> /Users/jankatins/.zinit/plugins/tldr-completion---gh-r/_tldr

[14:34:22] λ  ls -la /Users/jankatins/.zinit/plugins/tldr-completion---gh-r/_tldr
ls: /Users/jankatins/.zinit/plugins/tldr-completion---gh-r/_tldr: No such file or directory

[14:34:37] [1] ✗  zinit delete tldr-completion/gh-r -y
No such (plugin or snippet): tldr-completion/gh-r.
```

-> Will adjust the test but I can also imagine that there is some better way to handle a `as"completion"` which does not install a completion.

jankatins:
Test seems to work, please have a look...

vladdoster:
The tests can work, but this change isn't really tested, no?

jankatins:
The problem it's not tested right now neither: this PR shows that a lot of ICEs (=all exclamation mark based ones) do not work when zinit runs non-interactively. Given that all the test run non-interactively and no tests were failing before, all these ICEs are not tested properly. 

So the current situation is IMO better, as the two tests which had do be adjusted now at least show that the ICEs actually run. But, yes, all these ICEs which are implemented via hooks need better checks.

jankatins:
I added 4 tests for cp/mv/atclone/make which just basically test that the ices are actually run. I pushed only this commit in #265 to show that the tests fail on main without the fix of this branch, but apparently zunit always uses fail-fast (at least I couldn't make it run all tests locally, neither) :-( 

https://github.com/zdharma-continuum/zinit/runs/6368397193?check_suite_focus=true shows that at least the first test is failing on main.


On local I tried all 4 tests by deleting the test one by one from the top:

```
'/home/jan/.zinit/bin/tests/_support/zunit/plugins/test---mv/mv.md' does not exist or is not a file
[...]
'/home/jan/.zinit/bin/tests/_support/zunit/plugins/test---cp/cp.md' does not exist or is not a file
[...]
'/home/jan/.zinit/bin/tests/_support/zunit/plugins/test---atclone/atclone.md' does not exist or is not a file
[...]
'/home/jan/.zinit/bin/tests/_support/zunit/plugins/test---make/whatever' does not exist or is not a file
```

@vladdoster Anything else I should do in this PR?

jankatins:
Waiting for https://github.com/zdharma-continuum/zinit/pull/269 to land (previous runs were all green)

jankatins:
Rebased and green

tests/gh-r.zunit
```diff
@@ -145,7 +145,7 @@
   $dive --version; assert $state equals 0
 }
 @test 'docker-buildx' { # A monitor of resources
-  run zinit for as'completions' atclone'buildx* completion zsh > _buildx' lbin'!buildx-* -> buildx' @docker/buildx
+  run zinit for as'completions' atclone'./buildx* completion zsh > _buildx' lbin'!buildx-* -> buildx' @docker/buildx
```
vladdoster, line 148: Un-related, please remove

tests/ices.zunit
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env zunit
+
+@setup {
+  ZPLUGINS=$ZINIT[PLUGINS_DIR]
+
+  function _zunit_assert_not_exists() {
```
vladdoster, line 6: I think this assertion exists?

Screen Shot 2022-05-22 at 12 05 53

tests/ices.zunit
```diff
@@ -0,0 +1,52 @@
...
+@test 'atclone' {
+  run zinit as"null" id-as"test/atclone" atclone"mv readme.md atclone.md" for zdharma-continuum/null
+  assert $state equals 0
+  assert "$ZPLUGINS/test---atclone/atclone.md" is_file
+  assert "$ZPLUGINS/test---atclone/atclone.md" is_readable
+  assert "$ZPLUGINS/test---atclone/readme.md" not_exists
+}
+
+@test 'make' {
+  run zinit as"null" id-as"test/make" atclone"printf 'all:\n\ttouch whatever\n' > Makefile" make"" for zdharma-continuum/null
```
vladdoster, line 49: I'm surprised this works. I'd expect it to fail as AFAIK it tries to run `make install`.

tests/ices.zunit
```diff
@@ -0,0 +1,52 @@
...
+
+    [[ ! -e "$filepath" ]] && return 0
+
+    echo "'$pathname' does exist"
+    exit 1
+  }
+
+}
+
+@test 'mv' {
```
vladdoster, line 24: Do these tests verify this fix?

vladdoster: As in, are these tests running interactively?

zinit.zsh
```diff
@@ -1174,6 +1174,10 @@ builtin setopt noaliases
 # FUNCTION: @zinit-register-hook. [[[
 # Registers the z-annex inside Zinit – i.e. an Zinit extension
 @zinit-register-hook() {
+    emulate -LR zsh
+    # nobanghist to disable adding backslash escaping to "!", which is used in a lot of hocks and behaves differently
```
vladdoster, line 1178:
```suggestion
    # Set 'nobanghist' option disables backslash escaping for "!"
```

How does it behave differently?

zinit.zsh
```diff
@@ -1174,6 +1174,10 @@ builtin setopt noaliases
...
+    # in interactive sessions (escapes !) and non-interactive ones (does not)
+    builtin setopt extendedglob warncreateglobal typesetsilent noshortloops nobanghist
```
vladdoster, line 1180:
```suggestion
    builtin setopt extendedglob nobanghist noshortloops typesetsilent warncreateglobal
```

Sorted

vladdoster (changes requested):

tests/gh-r.zunit
jankatins: It fails the unit tests without as before the atclone ice wasn't run in the tests. If I remove this the CI run will fail (again)

(same for two more "fix: *" commits)

tests/ices.zunit
jankatins: The assertion, as it is written, only checks if a file does exist and reports if it is not existing. I've neither found a way to reverse any assertion now would it make sense: the assertion would report a file missing (`echo "'$pathname' does not exist"` and then would get reversed.

tests/ices.zunit
jankatins: They are running non-interactively, as all zunit tests. I'm not aware of any way to run them interactively (which probably is because of my zsh skill) :-(

Given that the bug this should fix is about making zinit able to run in (non-interactive) scripts, this is exactly what is getting fixed.

tests/ices.zunit
jankatins: All three places run `        command make -C "$dir" ${(@s; ;)${make#\!\!}}` so no mention of `install` (as far as I understand this line)

tests/snippets.zunit
```diff
@@ -52,9 +52,12 @@
 }
 @test 'tldr-completion::gh-r' {
   artifact="$ZINIT[PLUGINS_DIR]/tldr-completion---gh-r"
-  run zinit for as"completion" from"gh-r" id-as'tldr-completion/gh-r' bpick"completions_zsh" @dbrgn/tealdeer
+  run zinit for as"completion" from"gh-r" id-as'tldr-completion/gh-r' bpick"completions_zsh" mv"completions_zsh -> _tldr_ghr" pick"_tldr_ghr" @dbrgn/tealdeer
```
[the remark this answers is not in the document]
jankatins: To be honest, no idea: given that this shows up as a CI failure when the fix is applied and the test does not actually check for the existence of the completion, I would argue that it documents the status quo. Should I open a new bug report about this.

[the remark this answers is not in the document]
jankatins: Here is the output of a zinit call on current main with the old call in an interactive session, showing that no completion got installed:

```
~/.zinit/bin on  main (07cde660) [?] [🐳: local]

[22:31:59] λ  zinit delete tldr-completion/gh-r -y

Done (action executed, exit code: 0)


~/.zinit/bin on  main (07cde660) [?] [🐳: local]

[22:32:07] λ  zinit for as"completion" from"gh-r" id-as'tldr-completion/gh-r' bpick"completions_zsh" @dbrgn/tealdeer

Downloading dbrgn/tealdeer… (at label: tldr-completion/gh-r…)
(Requesting `completions_zsh'…)
#################################################################################################################################################################################################################### 100.0%
ziextract: WARNING: didn't recognize the archive type of `completions_zsh'  }(no extraction has been done).%f%b
No files for compilation found.
Warning: ∞zinit-compile-plugin-hook hook returned with 1


~/.zinit/bin on  main (07cde660) [?] [🐳: local]

[22:32:15] [1] ✗  ls -la ~/.zinit/completions/ | grep tldr


~/.zinit/bin on  main (07cde660) [?] [🐳: local]

[22:32:21] [1] ✗
```

zinit.zsh
jankatins: Example:

```zsh
λ  cat excl-test
test=x\!y ; echo "${(q)test}"  

λ  source excl-test
x\!y

λ  zsh -c 'source excl-test'
x!y
```

(from a [comment in the bug report](https://github.com/zdharma-continuum/zinit/issues/199#issuecomment-1092191551))

jankatins: I hope to get the "how it behaves is differently" in the next line

```
# ...behaves differently 
# in interactive sessions (escapes !) and non-interactive ones (does not)
```

tests/gh-r.zunit
jankatins: Example session: 

```
[22:44:37] [1] ✗  zinit for as'completions' atclone'buildx* completion zsh > _buildx' lbin'!buildx-* -> buildx' @docker/buildx

Downloading docker/buildx…
Cloning into '/Users/jankatins/.zinit/plugins/docker---buildx'...
⠧ ███████████ OBJ: 100, PACK: 21227/21227, COMPR: 100%, REC: 100%, RES: 100%
No files for compilation found.
Warning: ∞zinit-compile-plugin-hook hook returned with 1
(eval):1: no matches found: buildx*
Warning: ∞zinit-atclone-hook hook returned with 1
linkbin annex: The lbin ice (`buildx-* -> buildx')didn't match any files
```

zinit.zsh
jankatins: Reworded it a bit for hopefully more readability

zinit.zsh
jankatins: applied

tests/ices.zunit
jankatins: I did run with this fix for the last weeks and redid my setup multiple times, so I'm fairly confident it will not break interactive sessions.

jankatins:
Now the tests fail with

```
FAIL: pokeget
  default-ice: stored the ices: lucid '' as null nocompletions '' nocompile '' from gh-r lbin !.


Downloading talwat/pokeget…

gh-r: failed to find the correct GitHub release asset to download, modify bpick-ICE (current bpick: *.tar.gz).

'1' is not equal to '0'
  
Revolver process could not be found
Execution halted after failure
```

For me, https://github.com/talwat/pokeget returns a 404, seems that repo got deleted/set to private?

-> Disabled the test in an additional commit.

jankatins:
rebased again onto master with the [talwat/pokeget](https://github.com/talwat/pokeget) fix

vladdoster:
@psprint Any chance I could get your input on this change?

jankatins:
rebased on top of current master with the xtrace changes -> adjusted a similar line in this PR accordingly, too.

jankatins:
Also added some more details to the intial description:

> This (problem which this PR fixes) actually happened in the zunit tests as they run non-interactively and therefore any ices based on exclamation mark hooks were not ran. I fixed the few places where (after the fix in this PR) the ice were actually ran and errored due to mis-specifications. I also added explicit checks for a few ices. -> before this PR, ices are basically not tested (and mostly not testable!):-(

vladdoster:
@jankatins I'm waiting on getting another set of eyes on this change. Sorry to drag this PR out for so long.

.github/workflows/tests.yaml
```diff
@@ -66,6 +66,9 @@ jobs:
       - name: 'clean up zunit & revolver'
         run: rm -rf revolver zunit
 
+      - name: "ices"
+        run: zunit run tests/ices.zunit
+
```
vladdoster, line 71: Nit: the tests are sorted alphabetically

jankatins: fixed

jankatins:
Fun, the previos push had  `buildin` instead of `builtin` and nevertheless a green build...

jankatins:
rebased but now the gh-r tests fail


No clue whats going on here:

```
[12:15:37] λ  zinit  as'null' from"gh-r" nocompile nocompletions lbin for @Xhofe/alist

Downloading Xhofe/alist…
gh-r: failed to find the correct GitHub release asset to download.
```

`zinit lbin from"gh-r" as"null" for @alist-org/alist` works -> the org of that repo changed.

Seems curl requests do not get redirected, webbrowser ones do...

@vladdoster Added a fix for this, do you want an extra PR to get it applied to main or will you cherry pick it from here?

jankatins:
Seems the github/hub failure in the previous pushed commit was just a flaky test (api down/failed?) -> It's green again by just repushing :-)

jankatins:
@pschmitt @vladdoster Any chance to get a review for this? From my standpoint, it's a bug: you cannot run zinit in non-interactive mode without this fix, as some ices are missing/not run. This also affects zunit tests as they also run in non-interactive mode.

I've now run my setup with this PR for over a month and redone it several times on both a mac and a linux system. I'm pretty sure the "interactive" part still works and the non-interactive one also works as demonstrated by the (new and fixed) tests :-)

tests/ices.zunit
jankatins: Just FYI: this is the wording of an example in the readme: from my reading, it's in line with the code, i.e. if `install` is not the default make target, you have explicitly supply the target:

```shell
# Scripts built at install (there's single default make target, "install",
# and it constructs scripts by `cat'ing a few files). The make'' ice could also be:
# `make"install PREFIX=$ZPFX"`, if "install" wouldn't be the only default target.
zinit ice as"program" pick"$ZPFX/bin/git-*" make"PREFIX=$ZPFX"
zinit light tj/git-extras
```

jankatins:
Added https://github.com/zdharma-continuum/zinit/pull/227/commits/f54325eb12e74eb8b3951a15004784d71265c0ca to fix the tests (probably a leftover from #311?)

jankatins:
rebased and removed the test fix again as it is now in main.

jankatins:
@vladdoster @pschmitt Any chance to get this reviewed and then hopefully merged before the style changes hit?

jankatins:
https://github.com/zdharma-continuum/zinit/issues/321 is probably also affected by this bug as it runs the installation in non-interactive mode (cc: @xbc5)

.gitignore
```diff
@@ -21,3 +21,5 @@ txt/
 zmodules/
 tests/_output/
 tests/_support/zunit/
+
+.idea
```
pschmitt, line 25:
```suggestion
.idea
```

vladdoster:
@jankatins Is this ready for re-review?

jankatins:
Yes but nothing substantial has changed since your last review, but i answered your questions and comments as good as i could.

jankatins:
Rebased on top of current main and force pushed.

vladdoster:
@jankatins Okay, I found a test for this.

Can you run the following interactivley? It currently does not work.

This works
```
zi light zdharma-continuum/zinit-annex-rust
```

but this doesn't:

```
zi rustup cargo'!exa -> ls' load for zdharma-continuum/null
```

So AFAIK, these changes should make the aforementioned command work now.

vladdoster:
I'm also really wanting @psprint to look over this. This has been why I've held off on merging this. No personal qualms with changeset -- this fixes a huge bug in Zinit functionality (and makes the behavior consistent) :^).

jankatins:
> Then run (this does not work interactively):
> So AFAIK, these changes should make the aforementioned command work now.

My understanding of this bug+fix is that it only changes behavior for non-interactive usages. So i think it will still not work.

Will test tomorrow/day after when I'm back at my laptop.

jankatins:
Results of the above commands:

``` 
[16:18:39] λ  zi rustup cargo'!exa -> ls' load for zdharma-continuum/null

Downloading zdharma-continuum/null…
Cloning into '/home/jan/.local/share/zinit/plugins/zdharma-continuum---null'...
⠙ ███████████ OBJ: 100, PACK: 8/8, COMPR: 100%, REC: 100%
No files for compilation found.
Warning: ∞zinit-compile-plugin-hook hook returned with 1
rust annex: Running the rustup installer...
warning: rustup should not be installed alongside Rust. Please uninstall your existing Rust first.
warning: If you are sure that you want both rustup and your already installed Rust
error: cannot install while Rust is installed
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-std'
info: installing component 'rustc'
  nightly-x86_64-unknown-linux-gnu installed - rustc 1.64.0-nightly (6dbae3ad1 2022-07-25)
Rust is installed now. Great!
rust annex: Installing the requested crates...
    Updating crates.io index
  Downloaded exa v0.10.1
  Downloaded 1 crate (136.6 KB) in 0.76s
  Installing exa v0.10.1
  Downloaded ansi_term v0.12.1
  Downloaded glob v0.3.0
  Downloaded idna v0.2.3
  Downloaded jobserver v0.1.24
  Downloaded lazy_static v1.4.0
  Downloaded matches v0.1.9
  Downloaded log v0.4.17
  Downloaded locale v0.2.2
  Downloaded percent-encoding v2.1.0
  Downloaded pkg-config v0.3.25
  Downloaded natord v1.0.9
  Downloaded number_prefix v0.4.0
  Downloaded scoped_threadpool v0.1.9
  Downloaded pad v0.1.6
  Downloaded term_grid v0.1.7
  Downloaded tinyvec_macros v0.1.0
  Downloaded tinyvec v1.6.0
  Downloaded unicode-bidi v0.3.8
  Downloaded unicode-normalization v0.1.21
  Downloaded term_size v0.3.2
  Downloaded unicode-width v0.1.9
  Downloaded url v2.2.2
  Downloaded users v0.11.0
  Downloaded zoneinfo_compiled v0.5.1
  Downloaded form_urlencoded v1.0.1
  Downloaded cfg-if v1.0.0
  Downloaded cc v1.0.73
  Downloaded byteorder v1.4.3
  Downloaded bitflags v1.3.2
  Downloaded num_cpus v1.13.1
  Downloaded datetime v0.5.2
  Downloaded git2 v0.13.25
  Downloaded libc v0.2.126
  Downloaded libz-sys v1.1.8
  Downloaded libgit2-sys v0.12.26+1.3.0
  Downloaded 35 crates (5.8 MB) in 4.08s (largest was `libz-sys` at 2.5 MB)
   Compiling libc v0.2.126
   Compiling pkg-config v0.3.25
   Compiling tinyvec_macros v0.1.0
   Compiling log v0.4.17
   Compiling matches v0.1.9
   Compiling unicode-width v0.1.9
   Compiling percent-encoding v2.1.0
   Compiling unicode-bidi v0.3.8
   Compiling cfg-if v1.0.0
   Compiling bitflags v1.3.2
   Compiling byteorder v1.4.3
   Compiling ansi_term v0.12.1
   Compiling scoped_threadpool v0.1.9
   Compiling natord v1.0.9
   Compiling lazy_static v1.4.0
   Compiling number_prefix v0.4.0
   Compiling glob v0.3.0
   Compiling tinyvec v1.6.0
   Compiling pad v0.1.6
   Compiling term_grid v0.1.7
   Compiling form_urlencoded v1.0.1
   Compiling unicode-normalization v0.1.21
   Compiling jobserver v0.1.24
   Compiling locale v0.2.2
   Compiling users v0.11.0
   Compiling term_size v0.3.2
   Compiling num_cpus v1.13.1
   Compiling cc v1.0.73
   Compiling datetime v0.5.2
   Compiling idna v0.2.3
   Compiling zoneinfo_compiled v0.5.1
   Compiling exa v0.10.1
   Compiling url v2.2.2
   Compiling libz-sys v1.1.8
   Compiling libgit2-sys v0.12.26+1.3.0
   Compiling git2 v0.13.25
    Finished release [optimized] target(s) in 2m 16s
  Installing /home/jan/.local/share/zinit/plugins/zdharma-continuum---null/bin/exa
   Installed package `exa v0.10.1` (executable `exa`)
rust annex: Created the ls shim.

[16:38:26] λ  which ls
/usr/bin/ls

[16:39:37] λ  ls -la .local/share/zinit/plugins/zdharma-continuum---null/bin/
total 204152
drwxr-xr-x  2 jan jan     4096 Jul 26 16:22 .
drwxr-xr-x  7 jan jan     4096 Jul 26 16:39 ..
-rwxr-xr-x 13 jan jan 15825920 Jul 26 16:19 cargo
-rwxr-xr-x 13 jan jan 15825920 Jul 26 16:19 cargo-clippy
-rwxr-xr-x 13 jan jan 15825920 Jul 26 16:19 cargo-fmt
-rwxr-xr-x 13 jan jan 15825920 Jul 26 16:19 cargo-miri
-rwxr-xr-x 13 jan jan 15825920 Jul 26 16:19 clippy-driver
-rwxr-xr-x  1 jan jan  3290968 Jul 26 16:22 exa
-rwxr-xr-x 13 jan jan 15825920 Jul 26 16:19 rls
-rwxr-xr-x 13 jan jan 15825920 Jul 26 16:19 rustc
-rwxr-xr-x 13 jan jan 15825920 Jul 26 16:19 rustdoc
-rwxr-xr-x 13 jan jan 15825920 Jul 26 16:19 rustfmt
-rwxr-xr-x 13 jan jan 15825920 Jul 26 16:19 rust-gdb
-rwxr-xr-x 13 jan jan 15825920 Jul 26 16:19 rust-gdbgui
-rwxr-xr-x 13 jan jan 15825920 Jul 26 16:19 rust-lldb
-rwxr-xr-x 13 jan jan 15825920 Jul 26 16:19 rustup


[16:40:37] λ  ls -la .local/share/zinit/polaris/bin/ls
-rwxr-xr-x 1 jan jan 436 Jul 26 16:22 .local/share/zinit/polaris/bin/ls

[16:40:54] λ  cat .local/share/zinit/polaris/bin/ls
#!/usr/bin/env zsh

function ls {
    local bindir="/home/jan/.local/share/zinit/plugins/zdharma-continuum---null/bin"
    local -x PATH="/home/jan/.local/share/zinit/plugins/zdharma-continuum---null"/bin:"$PATH"
    local -x RUSTUP_HOME="/home/jan/.local/share/zinit/plugins/zdharma-continuum---null"/rustup CARGO_HOME="/home/jan/.local/share/zinit/plugins/zdharma-continuum---null"



    "$bindir"/"exa" "$@"

}

ls "$@"
``` 

But after opening a new shell:

``` 
[16:43:00] λ  which ls
/home/jan/.local/share/zinit/polaris/bin/ls
```

psprint:
It looks good to me.

vladdoster:
@psprint Thank you for a stamp of approval.

@jankatins Zunit tests failed due to network (API rate limiting). Kicked off failed zunit tests. Will merge this PR when they pass. 

@jankatins Thanks again for the work on this fix. :^)

jankatins:
@vladdoster / @psprint Woohoo!

Restarted the CI by doing a force push -> all green

jankatins:
Woohoo, thank you!

jankatins:
@psprint Summary is in the description, copied below to get it into the mail. The original issue is fixed via https://github.com/psprint/zinit/commit/e74365c88f8bae27d931d30d4de1d43bce4124d3 

Given that issue was undiscovered for ages, it seems a lot of ices are not covered by actual tests (e.g. test just adds an ice, but does not assure that the ice had an effect).

> Exclamation marks (per default) get escaped in interactive mode, but not in non-interactive modes. In this case it meant that zinit would not find any hooks which implements the ices when running in non-interactive mode (e.g. in a script). This in turn would NOT result in errors but would just silently not execute any of the ices which of course would be a bad thing(tm).

> This actually happened in the zunit tests as they run non-interactively and therefore any ices based on exclamation mark hooks were not ran. I fixed the few places where (after the fix in this PR) the ice were actually ran and errored due to mis-specifications. I also added explicit checks for a few ices. -> before this PR, ices are basically not tested (and the exclamation mark hook based ones not testable at all!) :-(

vladdoster:
@jankatins This broke patch-dl ice

jankatins:
It probably broke the annexes because I removed all `\\\\` thingies, but forgot the place where it was still escaped (in the annex hook registering)

zinit-install.zsh
```diff
@@ -451,9 +451,9 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
             # Store ices at clone of a plugin
             .zinit-store-ices "$local_path/._zinit" ICE "" "" "" ""
             reply=(
-                ${(on)ZINIT_EXTS2[(I)zinit hook:\\\!atclone-pre <->]}
-                ${(on)ZINIT_EXTS[(I)z-annex hook:\\\!atclone-<-> <->]}
```
jankatins, line 455: @vladdoster Probably this changed line broke https://github.com/zdharma-continuum/zinit-annex-patch-dl/issues/6

Issue #129, opened 2021-12-06
🐞 bug: zinit mv not renaming file during initial setup


Percee:
### Issue description

When starting zinit for the first time, using the direnv hook given in zinit wiki:

```
# Direnv hook into zsh
# https://zdharma.github.io/zinit/wiki/Direnv-explanation/
zinit from"gh-r" as"program" mv"direnv* -> direnv" \
    atclone'./direnv hook zsh > zhook.zsh' atpull'%atclone' \
    pick"direnv" src="zhook.zsh" for \
        direnv/direnv
```

The renaming (direnv* to direnv) does not happen:
`ziextract: Successfully extracted and assigned +x chmod to the file: `direnv.linux-amd64'.`

The file direnv.linux-amd64 isn't renamed to direnv.

If however I then rm -rf ~/.local/share/zinit/plugins/direnv---direnv and start zsh again, I get:

`ziextract: Successfully extracted and assigned +x chmod to the file: `direnv.linux-amd64'.
renamed 'direnv.linux-amd64' -> 'direnv'`

So it seems to only happen during the very first time zinit is setup.

### zinit config

```ZINIT_DIR="$HOME/.local/share/zinit/zinit.git"
if [[ ! -f "$ZINIT_DIR/zinit.zsh" ]]; then
    echo -e "y" | sh -c "$(curl -fsSL https://git.io/zinit-install)"
fi

source "$ZINIT_DIR/zinit.zsh"
autoload -Uz _zinit
(( ${+_comps} )) && _comps[zinit]=_zinit

# Load a few important annexes, without Turbo
# (this is currently required for annexes)
zinit light-mode for \
    zdharma-continuum/z-a-rust \
    zdharma-continuum/z-a-as-monitor \
    zdharma-continuum/z-a-patch-dl \
    zdharma-continuum/z-a-bin-gem-node

### End of Zinit's installer chunk

zinit ice wait lucid
zinit snippet OMZ::plugins/colored-man-pages
zinit ice wait lucid
zinit snippet OMZ::plugins/colorize
zinit ice wait lucid
zinit snippet OMZ::plugins/command-not-found
zinit ice wait lucid
zinit snippet OMZ::plugins/copydir
zinit ice wait lucid
zinit snippet OMZ::plugins/copyfile
zinit ice wait lucid
zinit snippet OMZ::plugins/cp
zinit ice wait lucid
zinit snippet OMZ::plugins/debian
zinit ice wait lucid
[...]

PR #241, opened 2022-05-01
fix: Don't error if $OPTS is not yet defined in .zinit-compinit call

jankatins:
This codepath was triggered when deleting the whole `~/.ziniti/plugins` dir: during recreation in `.zinit-prepare-home`, `.zinit-compinit &> /dev/null` would be called and the sourcing of `zinit.zsh` would error out with

```
+.zinit-compinit:2> [[.zinit-compinit:2: bad math expression: operand expected at `/Users/jan...'
 -n '' ]]
```

As the error was redirected to `/dev/null`, this never showed up. Any zinit install call during that session would not run any hooks as they were not yet registered. The next session would then be fine because the plugins dir would exist. 

## Motivation and Context

See https://github.com/zdharma-continuum/zinit/issues/129#issuecomment-1092111737 -> basically deleting the `~/.zinit/plugins` dir should result in complete recreation on the next run, but due to this bug, it had bad side effects.

Closes: #129 (probably, not tested on initial run, only after deleting the plugins dir)

## How Has This Been Tested?

Manually deleting the whole folder and starting the whole zinit setup from scratch

```zsh
rm -rf ~/.zinit/plugins/ && exec zsh
```

## Types of changes

## Checklist:

### DIFF
--- a/zinit-install.zsh
+++ b/zinit-install.zsh
@@ -601,7 +601,8 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
 #
 # No arguments.
 .zinit-compinit() {
-    [[ -n ${OPTS[opt_-p,--parallel]} && $1 != 1 ]] && return
+    # This might be called during sourcing when setting up the plugins dir, so check that OPTS is actually existing
+    [[ -n $OPTS && -n ${OPTS[opt_-p,--parallel]} && $1 != 1 ]] && return
 
     emulate -LR zsh
     builtin setopt nullglob extendedglob warncreateglobal typesetsilent


vladdoster:
@jankatins Awesome find and fix. I really appreciate the use of commit message using the prefix `fix:`!

vladdoster:
@pschmitt @jankatins The conditional should have used `-v`

```zsh
[[ -v SOME_VARIABLE ]]
```

pschmitt:
Pendantic but true, this is effectively the same though.

vladdoster:
Since it errors on operand and would result in smaller diff, is there a reason not to use 

```
(( expr ))
```

I see `[[ expr ]]` used elsewhere, but seems redundant to check the same variable twice.

@jankatins @pschmitt

vladdoster:
Regardless of the answer above, this brings up the issue of deleting zinit directories. 

As I've said before @jankatins, this is not how you should delete plugins.

Screen Shot 2022-05-02 at 01 55 51

Use `zinit delete --all`

I would assume this edge-case is handled in the delete command since it deletes subdirectories and not the actual `$ZINIT[PLUGINS_DIR]` directory. So, is this needed or just not using the proper method of deleting plugins?

pschmitt:
zsh man page :
> In other words,
>       [[ $var ]] is the same as [[ -n $var ]].  It is
       recommended that the second, explicit, form be used where
>        possible.

vladdoster:
> Regardless of the answer above, this brings up the issue of deleting zinit directories.
> 
> As I've said before @jankatins, this is not how you should delete plugins.
> 
> Screen Shot 2022-05-02 at 01 55 51
> 
> Use `zinit delete --all`
> 
> I would assume this edge-case is handled in the delete command since it deletes subdirectories and not the actual `$ZINIT[PLUGINS_DIR]` directory. So, is this needed or just not using the proper method of deleting plugins?

[My previous comment](https://github.com/zdharma-continuum/zinit/issues/199#issuecomment-1091017058)

vladdoster:
> zsh man page :
> 
> > In other words,
> > [[ $var ]] is the same as [[ -n $var ]].  It is
> > recommended that the second, explicit, form be used where
> > possible.

TIL :^)

vladdoster:
> > Regardless of the answer above, this brings up the issue of deleting zinit directories.
> > As I've said before @jankatins, this is not how you should delete plugins.
> > Screen Shot 2022-05-02 at 01 55 51
> > Use `zinit delete --all`
> > I would assume this edge-case is handled in the delete command since it deletes subdirectories and not the actual `$ZINIT[PLUGINS_DIR]` directory. So, is this needed or just not using the proper method of deleting plugins?
> 
> Link to [my previous comment](https://github.com/zdharma-continuum/zinit/issues/199#issuecomment-1091017058)
> 
> Screen Shot 2022-05-02 at 02 03 34

Indeed, it avoids this issue altogether


Screen Shot 2022-05-02 at 02 06 52

PR #372, opened 2022-09-12
use grep instead of the deprecated egrep

akliuxingyuan:
## Description 

replace `egrep` with `grep -E`

## Motivation and Context 

grep v3.8 show warning when use egrep

> egrep: warning: egrep is obsolescent; using grep -E

ref: https://www.phoronix.com/news/GNU-Grep-3.8-Stop-egrep-fgrep

## Usage examples 

```zsh

```

## Types of changes 

## Checklist: 

### DIFF
--- a/zinit-autoload.zsh
+++ b/zinit-autoload.zsh
@@ -1671,7 +1671,7 @@ ZINIT[EXTENDED_GLOB]=""
                       }
                   done
                   ICE=()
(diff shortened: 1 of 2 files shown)


pschmitt (approved):
lgtm

pschmitt:
but what about MacOS? 
Does the non-GNU grep even have an -E flag?

akliuxingyuan:
macOS seems to be using FreeBSD's grep, and FreeBSD's grep does offer `-E` flag
https://github.com/freebsd/freebsd-src/blob/373ffc62c158e52cde86a5b934ab4a51307f9f2e/usr.bin/grep/grep.c#L179

pschmitt:
Aight. Thanks!

PR #373, opened 2022-09-12
fix: Fix download from gh-r

yutkat:
## Description 

I tried to update plugins today (September 13, 2022) and the binary download from the GitHub release failed.
`gh-r: failed to find the correct GitHub release asset to download.`

It seems that the GitHub website has been updated and the method of parsing assets from HTML no longer works.
Check this command and video.
`curl https://github.com/BurntSushi/ripgrep/releases`
https://user-images.githubusercontent.com/8683947/189714315-fa1d28b0-755a-46f8-9239-5f16291df26b.mp4

I changed the process to get the list of released assets from the API.
This is a temporary patch and we may need to consider a few more things. However, I think it is better than not working.

## Reproduce steps 

.zshrc

```zsh
ZPLG_HOME="$HOME/.config/zinit"

if ! test -d "$ZPLG_HOME"; then
  mkdir -p "$ZPLG_HOME"
  chmod g-rwX "$ZPLG_HOME"
  git clone --depth 10 https://github.com/zdharma-continuum/zinit.git ${ZPLG_HOME}/bin
fi

typeset -gAH ZPLGM
ZPLGM[HOME_DIR]="${ZPLG_HOME}"
source "$ZPLG_HOME/bin/zinit.zsh"
autoload -Uz _zinit
(( ${+_comps} )) && _comps[zinit]=_zinit

zinit for \
    from'gh-r' \
    as'program' pick'ripgrep*/rg' \
  BurntSushi/ripgrep
```

Error

```zsh
gh-r: failed to find the correct GitHub release asset to download.
```

## Types of changes 

## Checklist: 

### DIFF
--- a/zinit-install.zsh
+++ b/zinit-install.zsh
@@ -1464,7 +1464,8 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
     if [[ -z $urlpart ]] {
         local tag_version=${ICE[ver]}
         if [[ -z $tag_version ]]; then
-            tag_version="$( { .zinit-download-file-stdout $site/latest || .zinit-download-file-stdout $site/latest 1; } 2>/dev/null | \
+            local releases_url=https://github.com/$user/$plugin/releases/latest
+            tag_version="$( { .zinit-download-file-stdout $releases_url || .zinit-download-file-stdout $releases_url 1; } 2>/dev/null | \
                               command grep -m1 -o 'href=./'$user'/'$plugin'/releases/tag/[^"]\+')"
             tag_version=${tag_version##*/}
         fi
--- a/zinit-install.zsh
+++ b/zinit-install.zsh
@@ -354,7 +354,13 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
 
     (
         if [[ $site = */releases ]] {
-            local url=$site/${ICE[ver]}
+            local tag_version=${ICE[ver]}
+            if [[ -z $tag_version ]]; then
+                tag_version="$( { .zinit-download-file-stdout $site/latest || .zinit-download-file-stdout $site/latest 1; } 2>/dev/null | \
+                                  command grep -m1 -o 'href=./'$user'/'$plugin'/releases/tag/[^"]\+')"
+                tag_version=${tag_version##*/}
+            fi
+            local url=$site/expanded_assets/$tag_version
 
             .zinit-get-latest-gh-r-url-part "$user" "$plugin" "$url" || return $?
 
@@ -1456,7 +1462,13 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
     local user=$1 plugin=$2 urlpart=$3
 
     if [[ -z $urlpart ]] {
-        local url=https://github.com/$user/$plugin/releases/$ICE[ver]
+        local tag_version=${ICE[ver]}
+        if [[ -z $tag_version ]]; then
+            tag_version="$( { .zinit-download-file-stdout $site/latest || .zinit-download-file-stdout $site/latest 1; } 2>/dev/null | \
+                              command grep -m1 -o 'href=./'$user'/'$plugin'/releases/tag/[^"]\+')"
+            tag_version=${tag_version##*/}
+        fi
+        local url=https://github.com/$user/$plugin/releases/expanded_assets/$tag_version
     } else {
         local url=https://$urlpart
     }


tylerw (commented):
I fixed this same issue locally with the following change, but your proposed fix is probably more robust: 
```
diff --git a/zinit-install.zsh b/zinit-install.zsh
index 531f5299..f0035372 100644
--- a/zinit-install.zsh
+++ b/zinit-install.zsh
@@ -354,7 +354,7 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
 
     (
         if [[ $site = */releases ]] {
-            local url=$site/${ICE[ver]}
+            local url=$site/expanded_assets/${ICE[ver]}
 
             .zinit-get-latest-gh-r-url-part "$user" "$plugin" "$url" || return $?
 
@@ -1456,7 +1456,7 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
     local user=$1 plugin=$2 urlpart=$3
 
     if [[ -z $urlpart ]] {
-        local url=https://github.com/$user/$plugin/releases/$ICE[ver]
+        local url=https://github.com/$user/$plugin/releases/expanded_assets/$ICE[ver]
     } else {
         local url=https://$urlpart
     }
```

yutkat:
@tylerw Oh! That's great. I didn't know that page existed.

Your idea looks better, so can you make a new PR instead of me?

yutkat:
It might not work well when ${ICE[ver]} is empty.

https://github.com/BurntSushi/ripgrep/releases/expanded_assets/

yutkat:
My current method works, but it causes a rate limit on the GitHub API, It might be better to get the latest tag somehow and fix this issue tylerw's method.

yutkat:
I updated the patch based on tylerw's idea.

numToStr:
Also, for some of the binary, zinit is downloading incompatible binary

image

My machine is x86_64 but zinit is downloading arm binaries. @yutkat Does this PR also fixes this issue?

crivotz:
I also have the same issue as you, and I use `bpick` to indicate which binary I want. This PR solves the problem that the `bpick` could not find the indicated version.

yutkat:
> My machine is x86_64 but zinit is downloading arm binaries. @yutkat Does this PR also fixes this issue?

I haven't changed the binary determination process so that's probably a different issue.

BTW, I'm also using `x86_64` and I was able to download `fd` properly.

numToStr:
> BTW, I'm also using x86_64 and I was able to download fd properly

@yutkat would you mind sharing your script for `fd`?

yutkat:
here
https://github.com/yutkat/dotfiles/blob/bea6bfd498e830ead373ae365d8edefec302983c/.config/zsh/rc/pluginlist.zsh#L257-L261

zinit-install.zsh
```diff
@@ -1456,7 +1462,13 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
     local user=$1 plugin=$2 urlpart=$3
 
     if [[ -z $urlpart ]] {
-        local url=https://github.com/$user/$plugin/releases/$ICE[ver]
+        local tag_version=${ICE[ver]}
+        if [[ -z $tag_version ]]; then
+            tag_version="$( { .zinit-download-file-stdout $site/latest || .zinit-download-file-stdout $site/latest 1; } 2>/dev/null | \
```
tylerw, line 1467:
```suggestion
            local releases_url=https://github.com/$user/$plugin/releases/latest
            tag_version="$( { .zinit-download-file-stdout $releases_url || .zinit-download-file-stdout $releases_url 1; } 2>/dev/null | \
```

tylerw: Thanks for working on this @yutkat! I checked out your PR locally and needed to make the above change in order for it work. It looks like `$site` isn't defined in scope at this point in the script.

yutkat: Thanks!

crivotz:
Just for info, today I tried a fresh zinit installation and the problem with du/rg no longer occurred

numToStr:
Yeah, I also just did `zinit update` everything seems to be working fine again, no errors and no downloading wrong binary.

mrroll:
I can confirm that this PR also fixes an issue I'm having with https://github.com/starship/starship.

My arch is under the More Assets button in the latest release like so:

image

`bpick` fails to find it and errors out when installing/updating. When I have this PR checked out, it downloads the correct archive.

yutkat:
@pschmitt @vladdoster Sorry for the mensch. But I think this is a pretty serious issue, so I hope you'll check this out as soon as possible.

jankatins:
Can confirm, it fixes the starship problem in https://github.com/zdharma-continuum/zinit/issues/374#issuecomment-1249511640

vladdoster:
Big thank you to @yutkat & @tylerw for fixing!!

Sorry for delay in merging :^/.

Cheers!

PR #380, opened 2022-09-28
fix: update gh-r unit tests

vladdoster:
Signed-off-by: Vladislav Doster <mvdoster@gmail.com>


vladdoster:
## Fixes issue

Zunit parsing would sometimes leak the Vim fold descriptions. AFAIK it did not affect test correctness but looked bad.

An example is shown below

### Before

Screenshot 2022-10-30 at 01 05 59

### After

Screenshot 2022-10-30 at 01 04 48

zinit-install.zsh
```diff
@@ -1451,6 +1454,32 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
 
     return $?
 } # ]]]
+# FUNCTION: .zinit-single-line [[[
+# Display cURL progress bar on a single line
+.zinit-single-line() {
```
vladdoster, line 1459: @psprint Does this seem reasonable? I am moving this code from its own file (i.e., `share/single-line.zsh`) into a function with the rest of the Zinit code.

psprint: I'm not sure, if it works then I guess it's fine.

PR #432, opened 2022-12-01
revert: "perf: Make scheduler be called in 10 seconds interval if no tasks"

vladdoster:
Reverts zdharma-continuum/zinit#429


vladdoster:
Closes #433

Issue #271, opened 2022-05-13
[bug]: Starship Install "no such file or directory: atclone./starship init zsh > init.zsh; ./starship completions zsh > _starship":


EmDash00:
### Describe the bug

I'm new to using zinit having just switched from OMZ yesterday. I was testing out having zinit manage the install for Starship; however, it appears to error.  appears that putting

### Steps to reproduce

The example on the repository's README.md states these lines should do the trick.

```
zinit ice as"command" from"gh-r" \ 
          atclone"./starship init zsh > init.zsh; ./starship completions zsh > _starship" \
          atpull"%atclone" src"init.zsh"
zinit light starship/starship
```

However, when I just pasted them in, I got the above error. There appears to be something wrong with the substitution perhaps.

### Expected behavior

I expect it to not error during the atclone step. 

### OS / Linux distribution

Linux-GNU / Ubuntu 20.04

### Zsh version

5.8

### Terminal emulator

Termite

EmDash00:
It appears raw copying the text leaves some improper whitespace characters. Deleting them and reinserting tabs or spaces seems to solve the problem.

vladdoster:
Ah, glad you got it figured out. Indeed that was the issue, which you can see below.

<img width="1100" alt="Screen Shot 2022-05-14 at 03 49 03" src="https://user-images.githubusercontent.com/10052309/168418528-ef04c5c3-3a89-49c6-a39f-91716a1dd0cc.png">

Additionally, here is how I would write the starship recipe

```zsh
zinit for as'completion' atclone"
      ./starship init zsh > init.zsh
      ./starship completions zsh > _starship" \
    from"gh-r" nocompile sbin src"init.zsh" \
  starship/starship
```

vladdoster:
zinit ice as"command" from"gh-r" \
    atclone"./starship init zsh > init.zsh; ./starship completions zsh > _starship" \
    atpull"%atclone" src"init.zsh"
zinit light starship/starship

vladdoster:
The benefit being that you avoid adding starship to your path, are more explicit about the completion, and remove ices (`atpull`) that aren't needed.

jankatins:
> remove ices (atpull) that aren't needed

Why is that not needed? If starship changes the init code, I want to get the updated code on update and I though atpull would run on update, not atclone? Or is that different for gh-r?

PR #480, opened 2023-02-24
fix: the starship example does not work properly

xqm32:
The error may be caused by the space after the backslash, and `sh` requires that the character after the backslash should be `\n` to be regarded as a continuation line

> A backslash preserves the literal meaning of the following	character, with the exception of the newline character	(`\n'). A backslash preceding a newline is treated as a line continuation.

refer to https://man.freebsd.org/cgi/man.cgi?query=sh

## Description 
Fixed wrong starship example in `README.md`

## Motivation and Context 
This PR addresses the unfixed issues raised by https://github.com/zdharma-continuum/zinit/issues/271, and makes the starship example work fine

## Related Issue(s) 
https://github.com/zdharma-continuum/zinit/issues/271

## Usage examples 
Same as the original usage

## Types of changes 

## Checklist: 

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,9 +210,12 @@ zinit ice pick"async.zsh" src"pure.zsh" # with zsh-async library that's bundled 
 zinit light sindresorhus/pure
 
 # Load starship theme
-zinit ice as"command" from"gh-r" \ # `starship` binary as command, from github release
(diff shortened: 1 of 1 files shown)


vladdoster:
@xqm32,

thanks for catching this. A dum question, but did you test this?

vladdoster:
@xqm32,

Thanks for confirming. Merged!

PR #491, opened 2023-03-27
feat: zinit completion improvements

vladdoster:
## Description 

- add all subcommand short & long flags
- handle mutually exclusive options. For example, a user shouldn't be able to select the `--all` flag and a plugin in `zi delete` but should be able to select the `--yes` or `--quiet`.
- handle the zinit limitation of only taking one plugin in various commands (i.e., delete, uncompiled).

Examples:
- Don't show users plugins in the `zi delete` when they have already selected one to avoid an error.
- Only let the user select one flag option in `zinit time`.

Screenshot 2023-03-27 at 04 33 16
Screenshot 2023-03-27 at 04 33 44
Screenshot 2023-03-27 at 04 33 57

## Motivation and Context 

Improve ease of use.

## Types of changes 

- [X] New feature (non-breaking change which adds functionality)

## Checklist: 

### DIFF
--- a/_zinit
+++ b/_zinit
@@ -1,344 +1,377 @@
 #compdef zinit
+#autoload
+#
+# Zinit ZSH completion function
+# Copyright (c) 2016-2020 Sebastian Gniazdowski
+# Copyright (c) 2016-2023 zdharma-continuum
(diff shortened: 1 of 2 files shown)


pschmitt (approved):
Works great!

PR #497, opened 2023-04-10
feat: refactor ls, list, and loaded commands

vladdoster:
## Description

Issues addressed:

1. Confusing and unintuitive command names

  Currently, two commands `loaded` and `list` will list installed
  plugins and a single command `ls` lists installed snippets. The names
  don't make much sense or give context to their behavior, nor does
  having two commands with identical behavior.

This PR removes the three commands in favor of two new commands `snippets` and `plugins`. The names align with the action they perform.

2. Uninformative output

  Current output is hard to grok.

This PR cleans up the output format and uses color to help convey information.

3. Incorrect

  Output does not account for various situations with the only way to
  fix is to reload Zsh session which would not always fix it. See images
  below for the aforementioned situations.
  
Note: PR changes on left, Zinit main branch on right

Screenshot 2023-04-09 at 19 03 11

Screenshot 2023-04-09 at 19 09 44

## Types of changes 

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] Documentation change

## Checklist: 

- [X] I have updated the documentation accordingly.


vladdoster:
This bug was the catalyst for this PR. 

This could be viewed from two angles.

1. Yes, even though the source code is _"deleted"_, the session retains functionality *for the most part* (e.g., gh-r are gone but zsh plugin still works) due to (shooting from the hip here) that the underlying C code is still in memory. So does reporting no changes after deleting wrong?
2. Once a plugin is deleted, it is considered not loaded even though the current shell retains the functionality.

... I need a 🚜 as I'm obviously in the weeds and being pedantic ... 


### Before

Screenshot 2023-04-10 at 03 04 53

### After

Screenshot 2023-04-10 at 03 14 01

The scope creep was strong with this one.

vladdoster:
You might have wondered why I use the particular style and colors I do. It is heavily inspired by brew.

For example:

Screenshot 2023-04-10 at 06 35 38

PR #500, opened 2023-04-14
feat: add lint gh action workflows

vladdoster:
Lint commit message and various file formats


vladdoster:
Closing due to the proposed workflows are already present.

PR #506, opened 2023-04-17
feat: expand gh-r zunit tests

vladdoster:
## Usage examples

```zsh

```

## Types of changes 

- [ ] Documentation change

## Checklist: 

### DIFF
--- a/tests/_support/gen-ghr-test
+++ b/tests/_support/gen-ghr-test
@@ -1,0 +1,50 @@
+#!/usr/bin/env zsh
+#
+# Easily generate gh-r Zunit tests.
+#
+# Usage:
+#   gen-ghr-test [<user/repository>...]
+#
+# Note: Tests might require slight adjustments to the lbin or id-as ice
+#
+(){
+  local flag_help
+  local -a usage=(
+      "gen-ghr-test [-h|--help]"
+      "gen-ghr-test [<project...>]"
+  )
+
+  # -D pulls parsed flags out of $@
+  # -E allows flags/args and positionals to be mixed, which we don't want in this example
+  # -F says fail if we find a flag that wasn't defined
+  # -M allows us to map option aliases (ie: h=flag_help -help=h)
+  # -K allows us to set default values without zparseopts overwriting them
+  # Remember that the first dash is automatically handled, so long options are -opt, not --opt
+
+  zmodload zsh/zutil
+  zparseopts -D -F -K -- \
+      {h,-help}=flag_help \
+      || return 1
+
+  [[ -z "$flag_help" ]] || { print -l $usage; return 0 }
+
+  (( !$# )) && {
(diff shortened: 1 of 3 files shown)


tests/_support/gen-ghr-test
```diff
@@ -0,0 +1,58 @@
...
+# Usage:
+#   gen-ghr-test \
+#       'ouch-org/ouch' \
+#       'sharkdp/pastel' \
+#       'Elkowar/pipr' \
+#       'jqnatividad/qsv' \
+#       'lotabout/rargs' \
+#       'nivekuil/rip' \
+#       'sirwart/ripsecrets' \
+#       'ismaelgv/rnr'
```
vladdoster, line 14:
```suggestion
#   gen-ghr-test [<user/repository>...]
```

tests/_support/gen-ghr-test
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env zsh
```
jankatins, line 1: At one point it might result in faster and more stable cycles to refactor the logic for selecting the right file into an extra function and build a corpus of api call result + annotated "wanted file for each OS" and ran that in a unit test. 

And then use this script only as a weekly test run or so :-)

vladdoster: @jankatins,

Sorry, I don't quite follow. This was a script I created not to have to copy-paste when I wanted to create multiple tests, and it isn't used in any test process. I just found it helpful and added it for others.

Are you suggesting creating mocks? If yes...

IMO, automated testing involves many different techniques, but one that shouldn’t be used is mocking. Mocks are a distraction at best and provide false confidence at worst.

The current test paradigm allows for an actual E2E test of Zinit other than deletion (which we should add) and ensures that annexes are correctly working.

Regarding gh-r tests, 75% is duplicate code and could be encapsulated in a function.

jankatins: No, not mocks, only testing the "find the right file to download" functionality based on real but saved responses.

From my understanding, the ghr test tests do the following:
- the ghr functionality, but the code path is very narrow despite having a lot of test cases -> Essentially it's a lot of duplicated tests: e.g. if two projects use the same github action to upload releases, there is no need to test both projects.
- The "file to download selection" logic, but in a fragile and slow manner: if a projects removes a file or renames the files, the tests might fail. To run all, you need to download, extract, and run all, which is really slow.

My suggestion was to refactor out the second part and make it narrowly scoped: get all the responses, dump them as examples into a test folder and annotate them: "with OS set to linux, Arch to x86, and the input file a, the filename should be ...". This test should essentially finish in a second or so -> If zinit then fails on a new/changed project, one could add another annotated input and run the test and then do fast iterations on the code to also catch that edge case.

The ghr test could then be reduced to a few reliable projects which actually exercise different code paths.

PR #521, opened 2023-05-15
fix: zsh impl of realpath handles old macOS BSD realpath

vladdoster:
## Usage examples

```zsh

```

## Types of changes 

- [ ] Documentation change

## Checklist: 


zinit-install.zsh
```diff
@@ -1198,13 +1219,8 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
     "${${(M)OPTS[opt_-q,--quiet]:#1}:+, skip the -q/--quiet option for more information}.{rst}"; retval=4; }
                 }
             } else {
-                if (( $+commands[realpath] )) {
-                    local rpv="$(realpath --version | head -n1 | sed -E 's/realpath (\(.*\))?//g')"
-                    if is-at-least 8.23 $rpv; then
-                        rel_url="$(realpath --relative-to="$local_dir/$dirname" "$url")" && \
-                            { url="$rel_url" }
-                    fi
-                }
```
vladdoster, line 1207: This is broken on macOS due to BSD `realpath` lacking feature parity with GNU `realpath`.

This change removes dependency via a zsh function.

pschmitt: Wouldn't it make more sense to use `readlink -f`?
Been using that on a bunch of repos I maintain where macos people complained about similar issues.

vladdoster: @pschmitt,

How can you compute the relative path between two files/directories with `readlink -f`?

pschmitt: Commented too quickly on that one. My mistake.

pschmitt (approved):

PR #531, opened 2023-06-05
ci: only build containers on main branch

pschmitt:
## Description

The containers should only be built when pushing to main.

## Types of changes 

- [ ] Documentation change

## Checklist: 

### DIFF
--- a/.github/workflows/containers.yaml
+++ b/.github/workflows/containers.yaml
@@ -1,6 +1,10 @@
 name: Container images
 
-on: [push, workflow_dispatch]
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/containers.yaml
```diff
@@ -1,6 +1,10 @@
 name: Container images
 
-on: [push, workflow_dispatch]
+on:
+  push:
+    branches:
+      - main
```
vladdoster, line 6:
```suggestion
      - main
    tags:        
      - v[0-9]+.[0-9]+.[0-9]+
```

vladdoster (changes requested):
Should include tags for releases

PR #618, opened 2024-01-09
feat: check for completion files

vladdoster:
## Usage examples

```zsh

```

## Types of changes 

- [ ] Documentation change

## Checklist: 


vladdoster:
Automatically detect completion files

Screenshot 2024-01-09 at 05 38 15

vladdoster:
@jankatins, thoughts?

zinit-install.zsh
```diff
@@ -528,78 +528,81 @@ builtin source "${ZINIT[BIN_DIR]}/zinit-side.zsh" || {
...
-
+    local abbrev_pspec=$REPLY 
     .zinit-exists-physically-message "$id_as" "" || return 1
-
-    # Symlink any completion files included in the plugin directory
-    typeset -a completions already_symlinked backup_comps
+    typeset -aU completions already_symlinked backup_comps
     local c cfile bkpfile
-    # The plugin == . is a semi-hack/trick to handle 'creinstall .' properly
+    nt () {
```
jankatins, line 552: whats `nt`?

jankatins:
To be honest, I'm fine with the "less magic" variant of renaming to `_something` instead of automagically figuring out a file which contains completions ("explicit is better than implicit" -> python zen :-)). I tried to figure out which line actually contains the trigger but and I think I found it (`**/(_.*|*.zsh*)` or so) and that looks fine...

What happens if two files end up getting symlinked under the same name?

tests/ices.zunit
```diff
@@ -39,27 +39,32 @@
...
+  run zinit as"null" id-as"test/ignored_completions" atclone"touch __init__.py; echo '#compdef valid' >| _valid" completions for zdharma-continuum/null
   assert $state equals 0
   assert "$ZPLUGINS/test---ignored_completions/_valid" is_file
   assert "$ZPLUGINS/test---ignored_completions/__init__.py" is_file
   assert "$ZINIT[COMPLETIONS_DIR]/_valid" is_file
   COMPS=( "$ZINIT[COMPLETIONS_DIR]"/_* )
   assert __init__.py is_not_value_in $COMPS
+  zinit delete --yes test/ignored_completions
+  run zinit cuninstall test/ignored_completions
+  assert $output contains 'No completions found for `test/ignored_completions'
```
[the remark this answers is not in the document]
vladdoster: @jankatins Good idea.

zinit-install.zsh
vladdoster: nt is just a random function name. It could be anything.

jankatins: Ok, stupid question: why not a speaking function name so one can actually know what it does by just reading the function name? reading the zinit code is hard enough even without random function names :-)

vladdoster: @jankatins,

lol, I just took the name from the [zsh docs](https://zsh.sourceforge.io/Doc/Release/Expansion.html#Expansion). But yes, I will rename it.

Screenshot 2024-01-18 at 1 55 51 PM

.github/workflows/tests.yaml
```diff
@@ -69,7 +69,10 @@ jobs:
           mkdir -p "$HOME/.local/bin" && echo "$HOME/.local/bin" >> $GITHUB_PATH
           git clone --depth 1 https://github.com/zdharma-continuum/zunit; cd ./zunit
           ./configure --prefix=$HOME/.local && make all install
-      
+
+      - name: "gh-r"
+        run: $HOME/.local/bin/zunit run --verbose tests/gh-r.zunit
```
pschmitt, line 74:
```suggestion
        run: zunit run --verbose tests/gh-r.zunit
```

PR #750, opened 2025-12-29
Feature: Add Native One-Time Installation for Local Plugins

alberti42:
#### Motivation and Problem

Currently, `zinit` treats local plugins (specified by an absolute path) differently from remote plugins. They lack a one-time "installation" phase, which means they are simply sourced "in-place" on every shell start.

This prevents the automatic, declarative setup of features like completions, which normally run during the `atclone` phase for remote plugins. As a result, users are forced to write complex `atload` hooks with manual flag files to simulate this one-time setup, making the configuration for local plugins unnecessarily verbose and complex.

For example, installing completions for a local plugin requires a workaround like this:

```zsh
# The OLD way: a complex, manual workaround
zinit ice atload"
  # Define a flag file in the current (plugin) directory.
  local flag_file='./.installed'

  # If the flag doesn't exist, run the one-time install.
  if [[ ! -f \$flag_file ]]; then
    # We are already inside the plugin's directory, so we use '.'
    zinit creinstall -q .
    
    # Create the flag file to prevent this from running again.
    touch \$flag_file
  fi
"
zinit light /path/to/my/local/plugin
```

This is not intuitive and goes against `zinit`'s goal of clean, declarative syntax.

#### Solution and Implementation

This pull request introduces a more integrated approach by treating a local plugin as a "pre-cloned" repository. It bridges the gap between remote and local plugin handling.

The implementation involves the following changes:

1.  **Detection in `.zinit-load()` (`zinit.zsh`):** The main load function is patched to detect when a new, unmanaged local plugin is being loaded. It then calls the main installation function, passing a special `local` type. A filter is included to prevent this logic from incorrectly triggering on `zinit`'s own internal use of local paths (e.g., loading from `$ZINIT[BIN_DIR]`).

2.  **Symlinking in `.zinit-setup-plugin-dir()` (`zinit-install.zsh`):** The main installation function is patched with a new branch to handle the `local` type. Instead of cloning or downloading, it creates a **symlink** from the user's local plugin directory into `zinit`'s managed `$ZINIT[PLUGINS_DIR]`.

3.  **Unified Post-Install Hooks:** After the symlink is created, the function proceeds to run **all the standard post-installation logic** that is normally reserved for remote plugins. This includes the automatic scan for completions triggered by the `completions` ice, running `make` hooks, etc.

4.  **Improved Logging:** The installation message has been updated to say `Installing local plugin...` instead of the confusing `Downloading %//...` when this new logic is triggered.

#### Benefits and User Experience

This change significantly improves the user experience for managing local plugins.

*   **Declarative Syntax:** Users can now use the simple, declarative `completions` ice for local plugins, just as they would for remote ones.
*   **Consistency:** Local plugins now behave almost identically to remote plugins during their initial setup.
*   **Simplified Configuration:** The need for manual `atload` workarounds is eliminated.
*   **Full Feature Parity:** Other installation-time hooks (`make`, `compile`, etc.) will now also work as expected for local plugins on their first run.

With this patch, the complex workaround from before becomes a simple, clean, and idiomatic `zinit` command:

```zsh
# The NEW way: clean and declarative
zinit ice id-as:'local/my-local-plugin' completions
zinit light /path/to/my/local/plugin
```

#### A Note on Update Tracking

A key design consideration is update tracking. Since `zinit` does not own the source of the local plugin (it only links to it), it cannot manage its updates via `zinit update`. The responsibility for keeping the local plugin's source directory up-to-date remains with the user. This is an expected and reasonable trade-off for the flexibility of managing plugins from one's own dotfiles or local projects.

### AI Assistance Disclaimer

This pull request, including the implementation of the patch and the descriptive text, was developed with the assistance of an AI model (Google's Gemini 2.5). The process was guided by a human developer who provided the specific logic, identified the target locations for surgical code changes, and directed the content and structure of this note. The final code and functionality have been thoroughly tested and validated by humans on macOS (Apple Silicon) and Linux (Ubuntu 22.04) to ensure correctness and stability.

### DIFF
--- a/.github/workflows/labeler.yaml
+++ b/.github/workflows/labeler.yaml
@@ -1,10 +1,10 @@
-name: Labeler
-
+name: label
 on:
   pull_request_target: # https://securitylab.github.com/research/github-actions-preventing-pwn-requests
 
 jobs:
   label-pull-requests:
+    name: "pull requests"
 
     permissions:
       contents: read
--- a/.github/workflows/linting.yaml
+++ b/.github/workflows/linting.yaml
@@ -1,4 +1,4 @@
-name: Lint
+name: lint
 
 on:
   pull_request:
@@ -11,7 +11,7 @@ on:
 
 jobs:
   mdlint:
-    name: Markdown Lint
+    name: markdown
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
@@ -40,13 +40,13 @@ jobs:
             -- *.md **/*.md
 
   zshlint:
-    name: ZSH Lint
+    name: zsh
     runs-on: ubuntu-latest
 
     strategy:
       fail-fast: false
       matrix:
-        task: [zsh-noexec, zsh-zcompile]
+        task: [noexec, zcompile]
 
     steps:
       - uses: actions/checkout@v6
@@ -63,11 +63,11 @@ jobs:
       - name: "run lint (${{ matrix.task }})"
         run: |
           case "${{ matrix.task }}" in
-            zsh-noexec)
+            noexec)
               find . -name '*.zsh' -type f -print0 \
                 | xargs -0 -n1 -P4 zsh -n
               ;;
-            zsh-zcompile)
+            zcompile)
               find . -name '*.zsh' -type f \
                 -exec zsh -fc "zcompile {}" \;
               ;;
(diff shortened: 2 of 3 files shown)


vladdoster:
@alberti42,

Sorry for delay in reviewing this PR. I will have a look at it and provide feedback sometime this week.

PR #751, opened 2025-12-30
Fix(self-update): Honor current branch instead of hardcoding 'main'

alberti42:
**Description:**

Hi all,

This pull request fixes an issue where `zinit self-update` fails for users who have cloned Zinit on a branch other than `main`.

#### **The Problem**

Currently, the `.zinit-self-update` function hardcodes the `main` branch and `origin/HEAD` in its git commands. If a user installs Zinit from a fork or a custom branch (e.g., a feature branch named `integrated`), running `zinit self-update` results in errors like:

```
fatal: ambiguous argument '..origin/HEAD': unknown revision or path not in the working tree.
...
fatal: Not possible to fast-forward, aborting.
```

This prevents users with non-standard setups from using the built-in update mechanism.

#### **The Solution**

This PR refactors the update logic to be branch-agnostic by making two key changes to the `.zinit-self-update` function in `zinit-autoload.zsh`:

1.  **Use the Configured Upstream:** The `git log` command now uses `..@{u}` instead of `..origin/HEAD`. The `@{u}` ref is shorthand for the configured upstream branch, making the log command dynamically adapt to the user's setup.
2.  **Generic `git pull`:** The `git pull` command has been modified to remove the hardcoded `origin main`. A plain `git pull` will automatically use the current branch's configured remote and upstream branch, which is the correct and expected behavior.

These changes ensure that `zinit self-update` respects the user's local Git configuration, allowing it to work seamlessly on any branch.

#### **How to Test**

1.  Clone Zinit and check out a non-`main` branch:
    ```sh
    git clone https://github.com/zdharma-continuum/zinit.git
    cd zinit
    git checkout -b my-test-branch
    # (Optional) Set up a remote fork to simulate the user's scenario
    ```
2.  Source the local `zinit.zsh` file from this branch.
3.  **On the `main` branch (without this PR):** Run `zinit self-update`. The command will fail.
4.  **On this PR's branch:** Run `zinit self-update`. The command should now succeed, pulling any new changes for `my-test-branch` without error.

This should improve the experience for developers and users who work with forks or custom branches. Thank you for your consideration!

#### **AI Assistance Disclaimer**

This pull request, including the implementation of the patch and the descriptive text, was developed with the assistance of an AI model (Google's Gemini). The process was guided by a human developer who provided the specific logic, identified the target locations for surgical code changes, and directed the content and structure of this note. The final code and functionality have been thoroughly tested and validated by humans on macOS (Apple Silicon) and Linux (Ubuntu 22.04) to ensure correctness and stability.

### DIFF
--- a/zinit-autoload.zsh
+++ b/zinit-autoload.zsh
@@ -1969,14 +1969,19 @@ print -- "\nAvailable ice-modifiers:\n\n${ice_order[*]}"
         [[ $1 = -q ]] && +zi-log "{pre}[self-update]{info} updating zinit repository{msg2}" \
 
         local nl=$'\n' escape=$'\x1b['
+        # Dynamically get the current branch name for logging and pulling
         local current_branch=$(command git -C $ZINIT[BIN_DIR] rev-parse --abbrev-ref HEAD)
-        # local current_branch='main'
+        # Warn if user is not on main (requested by maintainer)
+        if [[ -n $current_branch && $current_branch != main ]]; then
+            +zi-log "{pre}[self-update]{warn} non-{obj}main{warn} branch detected: {obj}${current_branch}{warn}. Self-update will pull from the branch’s configured upstream.{rst}"
+        fi
         local -a lines
         (
             builtin cd -q "$ZINIT[BIN_DIR]" \
             && +zi-log -n "{pre}[self-update]{info} fetching latest changes from {obj}$current_branch{info} branch$nl{rst}" \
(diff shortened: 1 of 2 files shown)


pschmitt (commented):
We should log a warning in case the current branch isn't main.

alberti42:
> We should log a warning in case the current branch isn't main.

I integrated the suggestion with commit 18bed78f8a2633753d78410627ee095878170128

PR #761, opened 2026-02-15
Fix double extraction when `extract` ice is used with `from'gh-r'`

alberti42:
### Problem

When using `extract'!'` (or `extract'!!'`) together with `from'gh-r'`, the downloaded archive was extracted **twice**:

1. First by the inline `ziextract` call in the gh-r download code
2. Then again by the `∞zinit-extract-hook`, which runs at `!atclone-pre`

On the second pass, the original archive no longer exists (it was deleted after the first extraction). The `--auto` detection then falls back to the `file` command, which identifies unrelated files (e.g., `.jar` files, which are technically zip archives) and attempts to extract them — producing errors and potentially corrupting the plugin directory.

Example error output:
```
[ziextract] detected a zip archive in the file lib/some-library-1.2.3.jar.
[ziextract] ERROR: the file `lib/some-library-1.2.3.jar' doesn't exist.
```

The same issue also affected the tarball download path.

### Root cause

The bug went unnoticed because `extract'!'` was typically used with cloned repositories (where only the hook runs), not with `from'gh-r'` releases (where both the inline call and the hook run).

### Fix

Guard the inline `ziextract` calls in both the gh-r and tarball download paths so they only run when the `extract` ice is **not** set. When the ice is set, the `∞zinit-extract-hook` handles extraction exclusively, avoiding the duplication.

### Documentation

Updated the `extract` ice description in `README.md` to document the previously undocumented prefix modifiers:

| Syntax | Behavior |
|---|---|
| `extract` | Auto-detect and extract archives |
| `extract'!'` | Extract and flatten one directory level |
| `extract'!!'` | Extract and flatten two directory levels |
| `extract'-'` | Extract but keep the archive file |
| `extract'!file.tar.gz'` | Extract a specific file (useful for cloned repos containing archives) |


alberti42:
@vladdoster Let me know if you need a minimal repro to test this bug and how it is fixed.

pschmitt (approved):

PR #762, opened 2026-02-15
Fix three bugs in `.zi-check-for-git-changes()` that broke `zinit self-update`

alberti42:
## Summary

Fix three bugs in `zinit self-update`.

### Bug 1: No fetch before change detection

`.zi-check-for-git-changes()` compared `HEAD` against the local tracking ref (`@{u}`) without fetching first, so the tracking ref was always stale. The `git fetch` that would have updated it lived inside `.zinit-self-update()`, in the block guarded by that very check — a chicken-and-egg problem.

**Result**: `zinit self-update` always reported "Already up-to-date." even when upstream had new commits.

### Bug 2: `git --work-tree` instead of `git -C`

`.zi-check-for-git-changes()` used `git --work-tree "$1"` for all its git commands. This flag overrides where git looks for tracked files, but does **not** change where git discovers the `.git` directory — that still happens by searching upward from the CWD. The correct flag is `git -C "$1"`, which changes directory first so that git finds the right `.git` and operates on the right repository.

The original author likely confused `--work-tree` with `-C`, interpreting it as "run git in this directory." In reality, `--work-tree` is meant for setups where the `.git` directory and working tree are in separate locations (e.g., bare repos with external worktrees, or dotfile management patterns). That's not the case for zinit, which is a standard clone.

**Result**: If the user's CWD was inside a different git repo, `rev-list` ran against the wrong repository, returning bogus commit counts and triggering unnecessary recompile/reload.

### Bug 3: `-q` flag never worked

Two separate issues prevented quiet mode from working:

1. **Dispatcher didn't forward arguments.** In `zinit.zsh`, the dispatcher for the `self-update` subcommand called `.zinit-self-update` without forwarding arguments:

    ```zsh
    (self-update)
        .zinit-self-update
        ;;
    ```

    Other dispatcher entries (e.g., `times`) use `"${@[2-correct,-1]}"` to forward arguments. Because `-q` was never passed through, `.zinit-self-update` always ran in verbose mode — all the `[[ $1 != -q ]]` checks always evaluated to true. (The only caller that actually passed `-q` was the internal `.zinit-self-update -q` call on line 3405.)

2. **`.zi-check-for-git-changes()` didn't accept `-q`.** Even if the dispatcher had forwarded `-q`, the log messages emitted by `.zi-check-for-git-changes()` (the "fetching latest changes" and "Already up-to-date." messages) had no quiet mode support.

**Result**: `zinit self-update -q` behaved identically to `zinit self-update`.

### Why bugs 2 and 3 went unnoticed

Bug 1 masked bug 2. Without a fetch, the local `@{u}` tracking ref was always equal to `HEAD`, so the `rev-list` comparison always returned `down=0`. The function always took the "no changes" exit path — making the `--work-tree` bug impossible to observe. Fixing the fetch (bug 1) made the `rev-list` comparison meaningful for the first time, which exposed bug 2.

Bug 3 went unnoticed because `.zi-check-for-git-changes()` never entered the code path where verbosity mattered — it always short-circuited with "Already up-to-date." due to bug 1. There was no visible difference between quiet and verbose when the function did nothing.

## Changes

- Replace all `git --work-tree "$1"` with `git -C "$1"` in `.zi-check-for-git-changes()` so git operates on the correct repository regardless of CWD
- Move `git fetch` into `.zi-check-for-git-changes()` so remote tracking refs are up-to-date before comparing
- Have `.zi-check-for-git-changes()` resolve the current branch name and return it via `$REPLY`, avoiding a duplicate `git rev-parse` call in `.zinit-self-update()`
- Move the "fetching latest changes" log message into `.zi-check-for-git-changes()` alongside the fetch
- Remove the now-redundant `git fetch` and log line from the `.zinit-self-update()` subshell
- Forward arguments from the dispatcher to `.zinit-self-update` in `zinit.zsh` using `"${@[2-correct,-1]}"`, matching the pattern used by other subcommands

# What was tested

- [x] Run `zinit self-update` from a different git repo directory when there are no upstream changes — reports "Already up-to-date." without recompiling
- [x] Run `zinit self-update` when upstream has new commits — fetches, display changelog, pull, and recompile
- [x] Run `zinit self-update` again immediately — reports "Already up-to-date."
- [x] Run `zinit self-update -q` — quiet mode still works correctly
- [x] Verified non-main branch warning still appears for forked repos on custom branches

### DIFF
--- a/tests/commands.zunit
+++ b/tests/commands.zunit
@@ -64,8 +64,13 @@
 
 @test 'self-update' {
   run zinit self-update
-  assert $output contains 'Already up to date.'
-  assert $state equals 0
+  assert $output matches 'Already up to date'; assert $state equals 0
+
+  run zinit self-update --help
+  assert $output matches '.*(HELP).*(self-update).*'; assert $state equals 1
+
+  run zinit self-update --quiet
+  assert $output is_empty; assert $state equals 0
 }
 @test 'set-debug' {
   ZINIT+=(DEBUG 'true')
--- a/tests/commands.zunit
+++ b/tests/commands.zunit
@@ -64,7 +64,7 @@
 
 @test 'self-update' {
   run zinit self-update
-  assert $output contains 'Already up-to-date.'
+  assert $output contains 'Already up to date.'
   assert $state equals 0
 }
 @test 'set-debug' {
--- a/tests/commands.zunit
+++ b/tests/commands.zunit
@@ -72,6 +72,27 @@
   run zinit self-update --quiet
   assert $output is_empty; assert $state equals 0
 }
+@test 'self-update --help / -h shows help and exits 1' {
+  for flag in '--help' '-h'; do
+    run zinit self-update $flag
+    assert $output contains 'self-update'
+    assert $output contains '--quiet'
+    assert $state equals 1
+  done
+}
+@test 'self-update --quiet / -q suppresses output' {
+  for flag in '--quiet' '-q'; do
+    run zinit self-update $flag
+    assert $output is_empty
+    assert $state equals 0
+  done
+}
+@test 'self-update rejects unknown flags' {
+  run zinit self-update --foo
+  assert $output contains 'Incorrect options given'
+  assert $output contains '--foo'
+  assert $state equals 1
+}
 @test 'set-debug' {
   ZINIT+=(DEBUG 'true')
   run +zi-log -n '{dbg} message'
--- a/zinit.zsh
+++ b/zinit.zsh
@@ -2953,7 +2953,6 @@ You can try to prepend {apo}${___q}{lhi}@{apo}'{error} to the ID if the last ice
                     .zinit-show-times "${@[2-correct,-1]}"
                     ;;
                 (self-update)
-                    .zinit-parse-opts "self-update" "$@"
                     .zinit-self-update
                     ;;
                 (unload)
--- a/zinit.zsh
+++ b/zinit.zsh
@@ -2243,8 +2243,7 @@ $match[7]}:-${ZINIT[__last-formatter-code]}}}:+}}}//←→}
         +zi-log "{b}{u-warn}ERROR{b-warn}:{rst}{msg2} Incorrect options given{ehi}:" \
                 "${(Mpj:$sep:)@:#-*}{rst}{msg2}. Allowed for the subcommand{ehi}:{rst}" \
                 "{apo}\`{cmd}$cmd{apo}\`{msg2} are{ehi}:{rst}" \
-                "{nl}{mmdsh} {opt}${allowed//\|/$sep2}{msg2}." \
-                "{nl}{…} Aborting.{rst}"
+                "{nl}  {opt}${allowed//\|/$sep2}{msg2}."
     } else {
         local -a cmds
         cmds=( load snippet update delete )
@@ -2636,6 +2635,9 @@ zinit() {
             if (( OPTS[opt_-h,--help] )); then
                 +zinit-prehelp-usage-message $cmd $___opt_map[$cmd] $@
                 return 1;
+            elif (( ${reply[(I)-*]} )); then
+                +zinit-prehelp-usage-message $cmd $___opt_map[$cmd] ${reply[@]}
+                return 1
             fi
         fi
     fi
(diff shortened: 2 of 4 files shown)


vladdoster (changes requested):
looks good other than my one change suggestion

vladdoster (dismissed):
@alberti42,

Users would expect `-q` and `--quiet` flags. 

Let's refactor this to use the current zinit command/option parser.

This has following benefits we get for free:

- self-update command usage via `-h`/`--help`
- global -q variable removing need to pass `-q` around and validate it
-attempt to follow current command convention

I can get it done today when I am free or I'm happy to help if you hit any issues with the refactor.

Refactor would be:

- [ ] Add [`self-update` here and you can reuse these value](https://github.com/zdharma-continuum/zinit/blob/main/zinit.zsh#L2626)

- [ ] Add it [here as well](https://github.com/zdharma-continuum/zinit/blob/34a18112e8866a7a7570b0ee1da44298677cc8d2/zinit.zsh#L2632)

- [ ] update your code and check for quiet flag via `(( !OPTS[opt_-q,--quiet] ))`

- [ ] Update [completion for `self-update`](https://github.com/zdharma-continuum/zinit/blob/34a18112e8866a7a7570b0ee1da44298677cc8d2/_zinit#L275) to have help and quiet flags like [cclear](https://github.com/zdharma-continuum/zinit/blob/34a18112e8866a7a7570b0ee1da44298677cc8d2/_zinit#L154)

vladdoster:
@alberti42 ,

LOL, I've had a PR open for 2 years addressing most of the `self-update` refactor in [fix(self-update): respect option flags](https://github.com/zdharma-continuum/zinit/pull/633). Whoops

alberti42:
I realized there are too many occurrences of `print -P "%F{xyz}...` overall. It makes no sense to patch up just a couple. It is more meaningful to understand how we want to have it and then handle it in a separate PR in a consistent manner.

pschmitt (approved):

pschmitt:
Yeah it's been very yolo and random.
I don't mind if you merge your own PRs, esp. when like this one here when there are no comments left.

vladdoster (commented):
Test flags to avoid regression

tests/commands.zunit
```diff
@@ -64,8 +64,34 @@
...
+    run zinit self-update $flag
+    assert $output is_empty
+    assert $state equals 0
+  done
+}
+@test 'self-update rejects unknown flags' {
+  run zinit self-update --foo
+  assert $output contains 'Incorrect options given'
+  assert $output contains '--foo'
+  assert $state equals 1
```
vladdoster:
```suggestion
@test 'self-update' {
  local cmd='self-update'
  run zinit $cmd
  assert $output matches 'Already up to date'; assert $state equals 0

  for flag in '-h' '--help'; do
    run zinit $cmd $flag
    assert $output matches ".*(HELP).*($cmd).*"; assert $state equals 1
  done

  for flag in '-q' '--quiet'; do
    run zinit $cmd $flag
    assert $output is_empty; assert $state equals 0
  done

  run zinit $cmd --invalid
  assert $output matches ".*(Incorrect options given).*(--invalid).*(Allowed for the subcommand.*$cmd).*"; assert $state equals 1
}
```

zinit.zsh
```diff
@@ -2623,18 +2622,22 @@ zinit() {
...
         if (( $@[(I)-*] || OPTS[opt_-h,--help] )); then
             .zinit-parse-opts "$cmd" "$@"
             if (( OPTS[opt_-h,--help] )); then
                 +zinit-prehelp-usage-message $cmd $___opt_map[$cmd] $@
                 return 1;
+            elif (( ${reply[(I)-*]} )); then
+                +zinit-prehelp-usage-message $cmd $___opt_map[$cmd] ${reply[@]}
+                return 1
             fi
         fi
```
vladdoster: @alberti42,

I think this can be refactored to be ... Idk why it checked twice 

```suggestion
        .zinit-parse-opts "$cmd" "$@"
        if (( OPTS[opt_-h,--help] )) || (( ${reply[(I)-*]} )); then
            +zinit-prehelp-usage-message $cmd $___opt_map[$cmd] ${reply[@]}
            return 1;
        fi
```

alberti42:
I have rebased the branch onto main, creating a clean linear history without unnecessary merge commits that had accumulated. `git diff` between the old and new branch tip shows no differences.

This PR fixes a long-standing bug that prevented `zinit self-update` from ever detecting upstream changes, meaning every user who ran it would silently get "Already up to date." regardless of reality. I find it a priority to get this merged.

If there are any remaining style or refactoring preferences, those can be addressed in a follow-up PR.

PR #787, opened 2026-07-02
fix(install): treat non-archive single file gracefully in ziextract.

vladdoster:
Change `ziextract` when no extractor case matches and no explicit type was passed (-z $ext), file is treated as plain single file with info message `'<file>' is not an archive — keeping it as a plain file.` plus chmod a+x. The existing exec-bit collection then picks it up, sets REPLY for --move, prints normal success. Error message retained for explicit bogus types (ziextract file bogus-type). One iteration mid-verify: dropped -- from chmod — BSD/macOS chmod rejects it (chmod: --: No such file or directory).

Why it broke: commit 1334994f (PR #771) made +x detection archive-exec-bit-only. Bare gh-r binaries arrive over HTTP with 644 and no extension => no extractor, no exec bit → error branch. Pre-#771, file(1) detection covered this case.

### Before

Screenshot 2026-07-02 at 06 41 11

### After

Screenshot 2026-07-02 at 06 41 23

## How Has This Been Tested?

Tests (tests/ices.zunit): two new — plain-file success path (asserts no error, message present, file executable) and bogus-explicit-type error
retention. Suite 16/16, including #771's ziextract-permissions. Regression-validated: plain-file test fails 15/16 with fix stashed.

## Types of changes 

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation change

## Checklist: 

- [X] I have added tests to cover my changes.

### DIFF
--- a/tests/ices.zunit
+++ b/tests/ices.zunit
@@ -82,6 +82,27 @@
   assert $output contains 'ERROR'
   assert $output contains "doesn't exist"
 }
+@test 'ziextract plain single file (not an archive)' {
+  # A bare gh-r-style release binary: no "unrecognized archive" error,
+  # the file is kept as-is and marked executable.
+  local workdir=$(mktemp -d)
+  printf '#!/bin/sh\necho jq\n' > "$workdir/jq-macos-arm64"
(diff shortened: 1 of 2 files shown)


vladdoster:
@pschmitt, any objections to merging this?

PR #812, opened 2026-08-28
fix: hide zinit's positional params from eval'd ices

### LINKED ISSUE #808: [bug]: Bare $@ inside a stored atclone ice expands to zinit-internal positional parameters on atpull'%atclone' replay, breaking the command
### What happened?

An `atclone` ice whose command embeds `$@` inside a double-quoted shell string (e.g. an awk program ending with `print "_fzf \"$@\""` to forward caller args) works when authored, but **breaks on update replay**.

The stored ice value (marker file) is re-executed by `.zinit-at-eval()` (`zinit-install.zsh:1858-1868`), which does:

```zsh
local atpull="$1" atclone="$2"
local cmd="$atpull"
[[ $atpull == "%atclone" ]] && cmd="$atclone"
eval "$cmd"
```

This function is called with positional parameters set (`$1` = the atpull spec, here `%atclone`, `$2` = the whole stored value — call sites at `zinit-install.zsh:2396` and `:2422`). During the `eval`, `$@` inside double quotes therefore expands to those **zinit-internal hook-runner arguments**, not to nothing/literal text. The expansion also shifts quote parity mid-string: in the stored sequence `\\\"$@\\\"`, the `\\` consumes its backslash pair and the following `"` closes the awk program early. awk then receives a truncated program and fails with `unterminated string`; the hook exits rc=1 and `zinit update` reports failure for the plugin.

Expected behavior: a stored ice value should be replayed byte-for-byte identically to how it ran at clone time. Shell expansions such as `$@` inside the user's ice must not resolve against zinit-internal positional parameters during replay.

Secondary observation: the two atpull-hook call sites pass the value inconsistently — quoted at line 2396, unquoted at line 2422 (`"$ICE[atclone]"` vs `$ICE[atclone]`).

### Steps to reproduce

Minimal standalone reproduction (no plugin needed) that mirrors `.zinit-at-eval()` semantics exactly — eval a double-quoted command containing `$@` from inside a function that has positional parameters:

```zsh
stored='awk "END{print \"_fzf \\\"$@\\\"\"}"'
replay() {
  local atpull="$1" atclone="$2"
  local cmd="$atpull"
  [[ $atpull == "%atclone" ]] && cmd="$atclone"
  mkdir -p /tmp/zinit-repro && cd /tmp/zinit-repro
  eval "$cmd"   # same as .zinit-at-eval()
}
set -- '%atclone' "$stored"   # mirrors .zinit-at-eval "$atpull" "$ICE[atclone]"
replay "$@"
```

Real-world shape that surfaced this: `atclone'… fzf --help | … | awk "… END{print \"_fzf \\\"\$@\\\"\"}" >| _fzf …'` + `atpull'%atclone' run-atpull`, then `zinit update <plugin>`. Escaping the expansion as `\$@` makes both clone-time and replay-time execution produce valid output (`_fzf "$@"`) — control case verified.

### Relevant output

Stored marker content (`~/.local/share/zinit/plugins/<id>/._zinit/atclone`) that triggered this on update:

```shell
fzf --help 2>&1 | grep -oE "^\s+--[^ =\[,]+" | sed "s/\s*//" | sort -u | awk "BEGIN{print \"#compdef fzf\"; print \"_fzf() {\"; print \"  local -a flags\"; print \"  flags=(\"} {print \"    \" \$0} END{print \"  )\"; print \"  _arguments -s 1::flags && return 0\"; print \"}\"; print \"_fzf \\\"$@\\\"\"}" >| _fzf && cp _fzf ~/.local/share/zinit/completions/_fzf
```

Replay result (standalone repro above):

```shell
awk: cmd. line:1: END{print "_fzf "%atclone
awk: cmd. line:1:           ^ unterminated string
replay rc=1
```

Note how `$@` expanded to `%atclone` (the first hook-runner argument) inside what should have been awk's literal output string. Control run with `\$@` escaped generates `_fzf "$@"` correctly.

### Screenshots and recordings

No screenshot captured.

### Operating System & Version

OS: linux-gnu | Vendor: pc | Machine: x86_64 | CPU: x86_64

### Zsh version

zsh 5.9.2 (x86_64-pc-linux-gnu)

### Terminal emulator

tmux-256color (`echo "${TERM}"` → `tmux-256color`); host terminal tmux

### If using WSL on Windows, which version of WSL

Not applicable (bare Linux)

### Additional context

Zinit version: v3.15.0-18-gfc234da3.

Workaround used: move generation out of inline ice strings into a helper script called by `atclone`/`atpull'%atclone'`, so no `$@` or nested quote escapes ever enter the stored value. A documentation note (or sanitization/warning when storing values containing `$@`/nested escapes) would prevent this class of failure.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

vladdoster:
## Description

Every user-provided ice body (`atclone`, `atpull`, `atinit`, `atload`, `make`, `configure`, `reset`, `ps-on-update`, `ps-on-unload`, `atdelete`, `wrap`, the `if`/`load`/`unload`/`wait` conditions, and the `zinit run` / `zinit pack` code paths) was eval'd inside a function whose positional parameters held zinit-internal values, so a bare `$@`, `$#`, or `$1` in the ice expanded to those internals. The worst case is `atpull'%atclone'` replay, where `$2` is the stored ice itself: the expansion breaks quote parity and the replayed command fails even though it worked at clone time.

The fix clears the positional parameters (`builtin set --`) after each function captures its arguments, following the existing precedent in `.zinit-load-snippet` and `.zinit-load-plugin`. Details:

- The dispatcher's `if''` checks eval inline with the positionals saved and restored around them, so a `return` in the condition still exits the whole `zinit` call and a `local` set by it stays visible to the plugin's later ices (legacy semantics preserved). As a side effect, a condition that runs its own `set --` can no longer corrupt the remaining plugin-spec list.
- The `trigger-load''` stub-defining anonymous function is invoked without `"$@"`, so definition-time expansions in the ice value no longer see the remaining plugin specs.
- Locals in functions that eval ices (`.zinit-run-delete-hooks`, `.zinit-confirm`, `.zinit-at-eval`) are `___`-prefixed so they no longer shadow user globals such as `$url` or `$plugin` inside the eval'd code.
- The duplicated `.zinit-at-eval` in `zinit-install.zsh` and `zinit-autoload.zsh` (whichever file was sourced last won) is deduplicated into `zinit-side.zsh`, and the `atpull` hook's `atclone` argument is quoted so an empty ice arrives as an empty `$2` instead of eliding.
- Bonus fix uncovered along the way: commit 72acc612 renamed the `wrap-track` ice to `wrap` in the ice list and call sites, but the handler kept the old function name and ice key, so every `wrap''` use failed with `command not found: .zinit-wrap-functions`. The rename is now completed (handler, ice key, README).
- `doc/zsdoc` regenerated from the final sources.

## Related Issue(s)

Closes #808

## Motivation and Context 

`$@`/`$#`/`$1` inside any ice silently expanded to zinit's internal arguments instead of behaving like an empty argument list, producing corrupted commands (#808) and, with common variable names like `$url`, silently wrong values inside `atdelete''` hooks. The `wrap''` ice has been entirely non-functional since its rename.

## Usage examples

```zsh
# Before: args.txt contains zinit's internal hook arguments; on atpull'%atclone'
# replay the expansion breaks quoting and the command fails.
# After: args.txt contains "args(0)[]" both at clone time and on replay.
zinit id-as"test/args" atclone'print -r -- "args($#)[$@]" > args.txt' \
    atpull'%atclone' run-atpull for zdharma-continuum/null

# `return` inside if'' still exits the whole zinit call (legacy semantics kept):
zinit for if'(( $+commands[fzf] )) || return 1' junegunn/fzf

# wrap'' works again:
zinit wrap'_p9k_precmd' for romkatv/powerlevel10k
```

## How Has This Been Tested?

- Full zunit suite on macOS (zsh 5.9.2), `gh-r.zunit` excluded as rate-limit bound: ices 27/27, ice-validation 43/43, commands 14/14, annexes 11/11, snippets 11/11, compile 6/6, config-zstyle 4/4, hook-handler 2/2, plugins (figlet/ncurses/lua-format subset) 3/3. `zsh -n` and `zcompile` clean on all five sources.
- New regression tests in `tests/ices.zunit`: `atclone`/`atpull'%atclone'`-replay/`if''` no longer see internal positional params, `return` in `if''` exits the whole zinit call, and `wrap''` installs its tracking wrapper.
- Each changed code path was additionally probed in isolated `zsh -f` sandboxes differentially against `main` (leak reproduced on `main`, absent on this branch; `if''` return/`local` semantics, atdelete hook argv, `zinit run` `-l`/multi-word behavior, and trigger-load stub bodies byte-identical to `main`).

## Types of changes 

- [ ] Documentation change

## Checklist: 

- [x] I have updated the documentation accordingly.


alberti42:
Hi @vladdoster, thanks for this PR! Looks good to me. A couple of remarks:

### Rebasing to ensure fresh docs

There is currently a conflict that needs resolving. Afterward it is important to do a rebase (not just resolve the conflict). The conflict is in a generated file. So after rebasing onto current `main`, we can regenerate the full docs, ensuring the committed docs reflect both the latest commits and the changes in this PR.

### Avoiding zinit internal variables from shadowing user variables

A side goal of this PR was:

> Locals in functions that eval ices (`.zinit-run-delete-hooks`, `.zinit-confirm`, `.zinit-at-eval`) are `___`-prefixed so they no longer shadow user globals such as `$url` or `$plugin` inside the eval'd code.

This is a very good improvement. However, checking more closely, there are many other such instances of variables dangerously shadowing user variables (`dir`, `hook`, `subtype`, `rc`, `quiet`, `tpe`, `type`, `url`). None of these is ever resolved dynamically -- they're only read by the same function that declares them -- so renaming them is a purely local edit. We could add the `___` prefix (or even better `___zinit_`) to these as well (but more on this later, because I think this could be done better in a separate PR).

To understand the mechanism of the shadowing, here is a short repro/demo:

```sh
export dir=/my/directory      # exported on purpose
export mydir=/my/directory    # same value, name zinit never uses

# A: single quotes -- $dir expanded LATER, by zinit
zinit as"null" id-as"A" atclone'print -r -- "A dir=[$dir]" >! r.txt' for zdharma-continuum/null

# B: double quotes -- $dir expanded NOW, by your shell
zinit as"null" id-as"B" atclone"print -r -- \"B dir=[$dir]\" >! r.txt" for zdharma-continuum/null

# C: single quotes, snippet starts a child process
zinit as"null" id-as"C" atclone'zsh -fc "print -r -- C dir=[\$dir]" >! r.txt' for zdharma-continuum/null

# D: single quotes, name zinit doesn't use
zinit as"null" id-as"D" atclone'print -r -- "D mydir=[$mydir]" >! r.txt' for zdharma-continuum/null

for id in A B C D; do cat ~/.local/share/zinit/plugins/$id/r.txt; done
```

Actual output:

```
A dir=[…/zinit/plugins/A]     ← zinit's value, not yours
B dir=[/my/directory]         ← yours
C dir=[]                      ← empty!
D mydir=[/my/directory]       ← yours
```

Case C is the ugly one: a local variable masks an exported one completely in zsh, so a child process started from a snippet sees `dir` as empty rather than either value.

One obvious caveat: a prefix improves the odds, it doesn't guarantee safety. Because of zsh's dynamic scoping, a user's snippet sees the locals of *every* function in the call chain, not just the innermost one. Instead, we could use `$ZINIT[...]`, such an associative array is the only thing in the codebase today that's genuinely collision-proof, since a user would have to name a variable literally `ZINIT` to clash. The associate array seems to me better than a prefix like `___` as it ensures protection.

### Discussion: the placeholder helper (suggest a follow-up PR)

There is a group of special variables that are resolved dynamically: `user`, `plugin`, `id_as`, `id_as_clean`, `save_url`, `local_path`, `local_dir`, `dirname`. These are better handled with a refactor in a separate PR.

`@zinit-substitute` is the function that expands the `%USER%` / `%PLUGIN%` / `%DIR%` placeholders inside ice values. It is not given the values to be replaced; it receives only the names of the ice variables it should rewrite, and then reads variables called `user`, `plugin`, `id_as` and so on out of whatever function happens to be above it in the call chain. So the variable name is the only thing connecting them.

It would be better if the helper could read `$ZINIT[user]` (or a `___zinit_`-prefixed name) instead. This would make these eight variables become as freely usable by users as everything else, and the shadowing problem would be completely resolved rather than mitigated.

My recommendation would be to merge this PR (after rebasing), and open a new refactoring PR, with the aim to protect user variables against being shadowed by internal zinit variables.

alberti42 (approved):
Thanks for this PR! My recommendation would be to merge this PR (after rebasing), and open a new refactoring PR, with the aim of protecting user variables against being shadowed by internal zinit variables (see my report in the PR discussion forum).

